PR #13, opened 2021-09-15
fix: adding in a custom error response so we get 200s instead of 404s

pogotc:
Apps that rely on client side routing will return a 404 when going to a page other than `/`, this is because that url doesn't correspond to a file in the S3 bucket. 

For example, in Otis going to `/talis/player` would return:

Screenshot 2021-09-15 at 10 21 19

whereas what we want is:

This PR adds in support for a new option called `isRoutedSpa` that's off by default, but when enabled it configures a custom error response in CloudFront so 404s from the backend are replaced with a 200. It also tweaks the S3 configuration to not set an `ErrorDocument` since this should only be needed for SPAs.

### DIFF
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -77,6 +77,14 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
             behaviors: [{ isDefaultBehavior: true }],
           },
         ],
+        errorConfigurations: [
+          {
+            errorCode: 404,
+            errorCachingMinTtl: 10,
+            responseCode: 200,
+            responsePagePath: "/index.html",
+          },
+        ],
       }
     );
     new cdk.CfnOutput(this, "DistributionId", {
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -53,7 +53,9 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
     this.s3Bucket = new s3.Bucket(this, "SiteBucket", {
       bucketName: siteDomain,
       websiteIndexDocument: props.websiteIndexDocument,
-      websiteErrorDocument: props.websiteErrorDocument,
+      websiteErrorDocument: props.isRoutedSpa
+        ? props.websiteErrorDocument
+        : undefined,
       publicReadAccess: true,
       removalPolicy: props.removalPolicy,
       autoDeleteObjects: props.removalPolicy === cdk.RemovalPolicy.DESTROY,
@@ -77,13 +79,15 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
             behaviors: [{ isDefaultBehavior: true }],
           },
         ],
-        errorConfigurations: [
-          {
-            errorCode: 404,
-            responseCode: 200,
-            responsePagePath: "/index.html",
-          },
-        ],
+        errorConfigurations: props.isRoutedSpa
+          ? [
+              {
+                errorCode: 404,
+                responseCode: 200,
+                responsePagePath: "/index.html",
+              },
+            ]
+          : undefined,
       }
     );
     new cdk.CfnOutput(this, "DistributionId", {
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -80,7 +80,6 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
         errorConfigurations: [
           {
             errorCode: 404,
-            errorCachingMinTtl: 10,
             responseCode: 200,
             responsePagePath: "/index.html",
           },
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -49,13 +49,18 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
       }
     );
 
+    let websiteErrorDocument: string | undefined = props.websiteErrorDocument
+    if (!websiteErrorDocument) { 
+      websiteErrorDocument = props.isRoutedSpa
+        ? props.websiteIndexDocument
+        : undefined
+    }
+
     // S3 bucket
     this.s3Bucket = new s3.Bucket(this, "SiteBucket", {
       bucketName: siteDomain,
       websiteIndexDocument: props.websiteIndexDocument,
-      websiteErrorDocument: props.isRoutedSpa
-        ? props.websiteErrorDocument
-        : undefined,
+      websiteErrorDocument,
       publicReadAccess: true,
       removalPolicy: props.removalPolicy,
       autoDeleteObjects: props.removalPolicy === cdk.RemovalPolicy.DESTROY,
@@ -84,7 +89,7 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
               {
                 errorCode: 404,
                 responseCode: 200,
-                responsePagePath: "/index.html",
+                responsePagePath: `/${props.websiteIndexDocument}`,
               },
             ]
           : undefined,
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -31,6 +31,10 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
   ) {
     super(scope, id);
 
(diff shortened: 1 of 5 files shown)


lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -77,6 +77,13 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
             behaviors: [{ isDefaultBehavior: true }],
           },
         ],
+        errorConfigurations: [
+          {
+            errorCode: 404,
+            responseCode: 200,
+            responsePagePath: "/index.html",
+          },
+        ],
```
pogotc, line 86: Not sure when you wouldn't want this enabled for an SPA with routing, but might be overkill for something that doesn't have routing, so maybe it should be toggleable through a new option in `CdnSiteHostingConstructProps`?

junglebarry: That suggestion sounds like the way to go - this can currently be used for static site hosting, too.

junglebarry:
Good idea. I saw this in your Catalyst config and did wonder if we should add it.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
junglebarry: I'm ultimately wondering if we just have an `isRoutedSpa` flag that controls both this and the error page in S3. I don't think we'll need the error page in S3 if this is configured.

pogotc: that sounds good to me, it makes sense to just have a simple flag for the calling app and have this library apply what we think are the best settings. I'll update the PR to do that.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -53,7 +53,9 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
     this.s3Bucket = new s3.Bucket(this, "SiteBucket", {
       bucketName: siteDomain,
       websiteIndexDocument: props.websiteIndexDocument,
-      websiteErrorDocument: props.websiteErrorDocument,
+      websiteErrorDocument: props.isRoutedSpa
+        ? props.websiteErrorDocument
+        : undefined,
```
pogotc, line 58: the `ErrorDocument` part of the S3 bucket is now no longer set unless you've opted into SPA routing

test/cdn-site-hosting/cdn-site-hosting-construct.test.ts
```diff
@@ -82,4 +81,48 @@ describe("CdnSiteHostingConstruct", () => {
...
+    test("configures an error document in S3", () => {
+      expectCDK(stack).to(
+        haveResourceLike("AWS::S3::Bucket", {
+          WebsiteConfiguration: {
+            ErrorDocument: "error.html",
+          },
+        })
+      );
+    });
+  });
```
pogotc, line 127: I've added this test here, I noticed in `cdn-site-hosting-with-dns-construct.test.ts` there's a lot of quite similar tests, should these tests be added there as well?

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -77,6 +79,15 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
             behaviors: [{ isDefaultBehavior: true }],
           },
         ],
+        errorConfigurations: props.isRoutedSpa
+          ? [
+              {
+                errorCode: 404,
+                responseCode: 200,
+                responsePagePath: "/index.html",
```
junglebarry, line 87: Does this depend on the value of `props.websiteIndexDocument`?

test/cdn-site-hosting/cdn-site-hosting-construct.test.ts
junglebarry: I added them in both places simply because if we decouple the two at a later point, we want regression coverage.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
junglebarry: I'm not sure about this. I think if someone specified `props.websiteErrorDocument` we should honour what they set; if they have omitted it and chosen `props.isRoutedSpa` we could make use `props.websiteIndexDocument` instead.

test/infra/cdn-site-hosting/cdn-site-hosting-construct.test.ts
```diff
@@ -82,4 +82,49 @@ describe("CdnSiteHostingConstruct", () => {
...
+      const app = new cdk.App();
+      stack = new cdk.Stack(app, "TestRoutedSPAStack", { env: testEnv });
+      new CdnSiteHostingConstruct(stack, "MyTestConstruct", {
+        certificateArn: fakeCertificateArn,
+        siteSubDomain: fakeSiteSubDomain,
+        domainName: fakeDomain,
+        removalPolicy: RemovalPolicy.DESTROY,
+        isRoutedSpa: true,
+        sources: [s3deploy.Source.asset("./")],
+        websiteErrorDocument: "",
```
junglebarry, line 99: This is not generally how you'd specify omitting this, is it? Does it suggest you need to make the property allow `undefined`?

junglebarry: Can you include a test where you provide a non-falsy value here, please?

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -77,6 +84,15 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
             behaviors: [{ isDefaultBehavior: true }],
           },
         ],
+        errorConfigurations: props.isRoutedSpa
+          ? [
+              {
+                errorCode: 404,
+                responseCode: 200,
+                responsePagePath: `/${props.websiteIndexDocument}`,
```
junglebarry, line 92: Is it possible to specify `websiteIndexDocument` with a leading slash? If so, what does it do?

junglebarry: (I'm not thinking you need to handle all sorts of complex paths; more wondering if there's any defence required.)

junglebarry (commented):
Looking good so far - a little tightening and a question (which hopefully results in a very simple answer, rather than code changes).

test/infra/cdn-site-hosting/cdn-site-hosting-construct.test.ts
maddocash: I did think about changing it but didn't because I assumed the the type would already allow undefined if CDK allowed it, but happy to update :)

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
maddocash: I just updated what was here before which was `/index.html` so I guess we could end up in a situation where someone sets `websiteIndexDocument` with a leading slash, I can add a guard for this.

test/infra/cdn-site-hosting/cdn-site-hosting-construct.test.ts
junglebarry: The type for the props is defined in this project, regardless what the underlying constructs support - you'll need to change the props definition to drop it, but that seems reasonable to me.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -31,6 +31,10 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
   ) {
     super(scope, id);
 
+    if (props.websiteIndexDocument[0] === "/") {
```
junglebarry, line 34: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith ?

junglebarry (commented):
The particular combination I'd like to see tested is one where `isRoutedSpa: true, websiteErrorDocument: "error.html"` - I don't think we have this covered anywhere, but it's important to understand what it's going to do.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
maddocash: TIL

maddocash:
> The particular combination I'd like to see tested is one where `isRoutedSpa: true, websiteErrorDocument: "error.html"` - I don't think we have this covered anywhere, but it's important to understand what it's going to do.

Test added: The behaviour is as you suggested earlier, if there is an errorDocument specified we use it. If not we check to see if it is an routed SPA

junglebarry (approved):
Thanks Ash! LGTM

PR #23, opened 2021-11-11
fix: changes to LambdaWorker and AuthenticatedAPI as we start to use them for real

malcyL:
https://github.com/talis/platform/issues/5074

This PR makes changes to the `authenticated-api` construct and `lambda-worker` construct now that we are trying to use them outside of the examples within the project.

### DIFF
--- a/lib/authenticated-api/authenticated-api.ts
+++ b/lib/authenticated-api/authenticated-api.ts
@@ -92,8 +92,9 @@ export class AuthenticatedApi extends cdk.Construct {
           functionName: `${props.prefix}${props.name}-${routeProps.name}`,
 
           entry: routeProps.lambdaProps.entry,
-          handler: routeProps.lambdaProps.handler,
           environment: routeProps.lambdaProps.environment,
+          handler: routeProps.lambdaProps.handler,
+          role: routeProps.lambdaProps.role,
 
           // Enforce the following properties
           awsSdkConnectionReuse: true,
--- a/lib/authenticated-api/authenticated-api.ts
+++ b/lib/authenticated-api/authenticated-api.ts
@@ -71,9 +71,9 @@ export class AuthenticatedApi extends cdk.Construct {
         awsSdkConnectionReuse: true,
         runtime: lambda.Runtime.NODEJS_14_X,
         timeout: cdk.Duration.minutes(2),
-        securityGroup: props.securityGroup,
-        vpc: props.vpc,
-        vpcSubnets: props.vpcSubnets,
+        /* securityGroup: props.securityGroup, */
+        /* vpc: props.vpc, */
+        /* vpcSubnets: props.vpcSubnets, */
       }
     );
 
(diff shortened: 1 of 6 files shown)


lib/authenticated-api/authenticated-api.ts
```diff
@@ -92,6 +92,7 @@ export class AuthenticatedApi extends cdk.Construct {
           functionName: `${props.prefix}${props.name}-${routeProps.name}`,
 
           entry: routeProps.lambdaProps.entry,
+          environment: routeProps.lambdaProps.environment,
```
malcyL, line 95: Need to be able to set environment variables on routes

lib/authenticated-api/authenticated-api.ts
```diff
@@ -104,6 +105,12 @@ export class AuthenticatedApi extends cdk.Construct {
         }
       );
 
+      if (routeProps.lambdaProps.policyStatements) {
+        for (const statement of routeProps.lambdaProps.policyStatements) {
+          routeLambda.role?.addToPolicy(statement);
+        }
+      }
```
malcyL, line 112: Pacing through a `role` didn't work well. It meant the client app had to setup everything automatically set on the role by the CDK. This was problematic.

Instead, pass in policy statements to be added to the role created by the CDK.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -134,9 +141,9 @@ export class AuthenticatedApi extends cdk.Construct {
       const durationMetric = routeLambda.metric("Duration");
       const durationAlarm = new cloudwatch.Alarm(
         this,
-        `${props.name}-${routeProps.name}-duration-alarm`,
+        `${props.prefix}${props.name}-${routeProps.name}-duration-alarm`,
```
malcyL, line 144: Add the prefix to the name of these alarms.

lib/lambda-worker/lambda-worker.ts
```diff
@@ -16,6 +16,17 @@ const MINIMUM_MEMORY_SIZE = 1024;
...
+  // should be placed on.
+  public lambdaQueueArn: string;
+
+  // The URL of the queue messages for this LambdaWorker to process
+  // should be placed on. If you have passed in an option subscrip topic,
+  // this queue will have automatically been subscribed to that topic.
+  // If you have not supplied a topic, then you will need to route
+  // messages to the queue at lambdaQueueUrl.
+  public lambdaQueueUrl: string;
+
```
malcyL, line 29: Expose the `lambdaQueueArn` and `lambdaQueueUrl`  so that we can chain workers, with one writing messages to the queue of the next.

I did consider exposing the complete SQS construct which would allow the user to read both of these properties. However, that would give the client full access to modifying the queue, perhaps in ways we don't want.

lib/lambda-worker/lambda-worker.ts
```diff
@@ -107,13 +119,25 @@ export class LambdaWorker extends cdk.Construct {
...
+      }
+    }
+
     // Add main queue and DLQ as event sources to the lambda
     // By default, the main queue is enabled and the DLQ is disabled
     lambdaWorker.addEventSource(
-      new eventSource.SqsEventSource(lambdaQueue, { enabled: true })
+      new eventSource.SqsEventSource(lambdaQueue, {
+        enabled: true,
+        batchSize: 1,
```
malcyL, line 133: Always set the batch size to one. We need to be able to replay individual failures.

ShaunParsons (approved):

PR #24, opened 2021-11-12
feat: plat-5187 Add container support to LambdaWorker construct

stephenrob:
Adds support to LambdaWorker for specifying a container image tag from an existing ECR Repository.

Adds `container-lambda-example` to show creating the ECR Repository, Building the docker image and then creating a basic lambda worker.


lib/lambda-worker/lambda-worker.ts
```diff
@@ -45,6 +47,12 @@ export class LambdaWorker extends cdk.Construct {
       );
     }
 
+    if (!this.isContainerLambda(props) && !this.isFunctionLambda(props)) {
+      throw new Error(
+        `Invalid lambdaProps only dockerImageTag/ecrRepositoryArn/ecrRepositoryName or handler/entry can be specified.`
+      );
+    }
```
mrliptontea, line 54: Hmm I think this check would throw when neither is specified, which I think we need such check. But another check is needed to ensure we don't have both set.

lib/lambda-worker/lambda-worker.ts
```diff
@@ -219,4 +206,89 @@ export class LambdaWorker extends cdk.Construct {
...
+        environment: props.lambdaProps.environment,
+        memorySize: props.lambdaProps.memorySize,
+        reservedConcurrentExecutions:
+          props.lambdaProps.reservedConcurrentExecutions,
+        retryAttempts: props.lambdaProps.retryAttempts,
+        securityGroup: props.lambdaProps.securityGroup,
+        timeout: props.lambdaProps.timeout,
+        vpc: props.lambdaProps.vpc,
+        vpcSubnets: props.lambdaProps.vpcSubnets,
+      });
```
mrliptontea, line 268: Could these be moved to inside factory functions, `return this.createContainerLambda(props);` and `return this.createFunctionLambda(props)`, in order to make `createLambdaFunction` shorter?

examples/container-lambda-worker/README.md
```diff
@@ -0,0 +1,14 @@
...
+The `cdk.json` file tells the CDK Toolkit how to execute your app.
+
+## Useful commands
+
+- `npm run build` compile typescript to js
+- `npm run watch` watch for changes and compile
+- `npm run test` perform the jest unit tests
+- `cdk deploy` deploy this stack to your default AWS account/region
+- `cdk diff` compare deployed stack with current state
+- `cdk synth` emits the synthesized CloudFormation template
```
malcyL, line 14: We should give some details here. And a diagram. See https://github.com/talis/talis-cdk-constructs/tree/main/examples/simple-lambda-worker

lib/lambda-worker/lambda-worker.ts
stephenrob: There is a test case for all being set and ensuring that the error is thrown, see the array/table of tests that are ran for these options.

mrliptontea: Ah sorry, I commented before looking at `isContainerLambda` and `isFunctionLambda` - all good here.

malcyL (approved):

PR #34, opened 2021-12-17
feat: pin CDK without fuzzy versioning

junglebarry:


package.json
```diff
@@ -43,30 +43,30 @@
...
+    "@aws-cdk/aws-route53-targets": "1.129.0",
+    "@aws-cdk/aws-s3": "1.129.0",
+    "@aws-cdk/aws-s3-deployment": "1.129.0",
+    "@aws-cdk/aws-sns": "1.129.0",
+    "@aws-cdk/aws-sns-subscriptions": "1.129.0",
+    "@aws-cdk/aws-sqs": "1.129.0",
+    "@aws-cdk/core": "1.129.0",
     "cdk-ecr-deployment": "0.0.80",
     "talis-node": "git://github.com/talis/talis-node.git#v0.0.3",
     "uuid": "^8.3.2"
```
malcyL, line 72: We usually try to pin `dependencies` but allow `dev-dependencies` to have `^` in platform projects.

Now we have decided to pin the `aws-cdk` - it seems even more unusual that this one `dependency`, `uuid` is left with a `^`.

Should this be pinned as well? 

I'm not going to not thumb this PR because of this though - I know how painful this has been.

malcyL (approved):

package.json
junglebarry: Good shout - done!

PR #37, opened 2022-02-03
feat: common cdk constructs

stephenrob:
Adds common base CDK constructs that are now reused across internal apps, brought into cdk-constructs as part of the LTI work.

### TalisCdkStackProps
Properties for any custom Talis CDK Stack, these are on top of the default CDK Stack props.

| Property | Type | Required | Description |
| -- | -- | -- | -- |
| vpcId | string | Optional | ID of the VPC to deploy stack components into |
| deploymentEnvironment | TalisDeploymentEnvironment | Required | The environment this stack is being deployed into |

### TalisCdkStack
A base stack on top of the default CDK Stack on which to build a custom Talis App CDK Stack, provides helper methods and variables for referencing commonly used components including environment based removal policies and vpc resolution.

```ts
const app = new cdk.App()
const props: TalisCdkStackProps = {
  deploymentEnvironment: TalisDeploymentEnvironment.TEST
}
const stack = new TalisCdkStack(app, 'my-stack', props)

// Get the removal policy for the stacks environment
// Returns a CDK Removal Policy
stack.getRemovalPolicyForTalisDeploymentEnvironment(props.deploymentEnvironment)
```

### TalisDeploymentEnvironment
Helper constants to allow us to map and reference our deployment environments

| Key | Description |
| -- | -- |
| DEVELOPMENT | Local development environment |
| TEST | Environments spun up for running tests |
| BUILD | Environments spun up for running builds |
| STAGING | Staging environment |
| PRODUCTION | Production Environment |

### TalisRegions
Helper constants to allow us to more easily map our product regions to AWS regions.

| Key | Description |
| -- | -- |
| CANADA | ca-central-1 |
| EU | eu-west-1 |

### Resource Prefixer Aspect
Defines a CDK Aspect that can be applied to a cdk stack to prefix all resources in the given stack with a given prefix. Useful for creating multiple CDK stacks in the same account but prefixing the resources for a given developer.

```ts
import { ResourcePrefixer } from "talis-cdk-constructs"

const app = new cdk.App()
const props: TalisCdkStackProps = {
  deploymentEnvironment: TalisDeploymentEnvironment.TEST
}
const stack = new TalisCdkStack(app, 'my-stack', props)

// Apply the aspect to the stack
// This will prefix all resources with 'dev-xx-'
// I.e. a dynamodb table called 'archive' in the stack will now be named 'dev-xx-archive'
Aspects.of(stack).add(new ResourcePrefixer('dev-xx-'))
```


lib/common/aspects/resource_prefixer/resource_prefixer.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.prefixers = new Map<
+      Constructor<CfnResource>,
+      Constructor<CfnResourcePrefixer>[]
+    >();
+
+    this.registerPrefixer(CfnTable, DynamoDbCfnTablePrefixer);
+  }
+
+  public visit(node: IConstruct): void {
+    // We only care about Cloudformation Resources so skip anything that isnot one
```
maddocash, line 26:
```suggestion
    // We only care about Cloudformation Resources so skip anything that is not one
```

test/fixtures/infra/cfn_table.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const CfnTableProperties: CfnTableProps = {
+  keySchema: [
+    {
+      attributeName: "attributeName",
+      keyType: "keyType",
+    },
+  ],
+
+  // the properties below are optional
```
maddocash, line 11: this file is a little confusing as you have this comment in a couple of places so I'm not sure what is optional and what isn't.
e.g. from this comment is `attributeDefinitions` optional but `globalSecondaryIndexes` isn't?

lib/common/aspects/resource_prefixer/resource_prefixer.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.prefix = prefix;
+    this.prefixers = new Map<
+      Constructor<CfnResource>,
+      Constructor<CfnResourcePrefixer>[]
+    >();
+
+    this.registerPrefixer(CfnTable, DynamoDbCfnTablePrefixer);
+  }
+
+  public visit(node: IConstruct): void {
```
maddocash, line 25: I'm not sure what this is for?

malcyL:
Instead of this being `lib/common` would this be better names `lib/stack` or `lib/talis-cdk-stack`.

I think what we are creating here is the pattern for how all of are stacks should be created, so having it names as a stack, which includes the additional things here, seems clearer to me than `common`.

malcyL:
The documentation in the PR above is good. 

We have created [examples](https://github.com/talis/talis-cdk-constructs/tree/main/examples) for other constructs though and I think we should follow that pattern for new constructs too. Can we create a quick example of using it, together with the documentation from the PR above.

stephenrob:
> The documentation in the PR above is good.
> 
> We have created [examples](https://github.com/talis/talis-cdk-constructs/tree/main/examples) for other constructs though and I think we should follow that pattern for new constructs too. Can we create a quick example of using it, together with the documentation from the PR above.

Not sure what examples for the common base parts would look like besides documentation, there's not exactly an example that can be spun up and have tests ran against.

test/fixtures/infra/cfn_table.ts
stephenrob: So this file is generated from the AWS docs of what a Cloudformation DynamoDB Table and just used as a test fixture

lib/common/aspects/resource_prefixer/resource_prefixer.ts
stephenrob: This is the method that needs to be implemented when using the `IAspect` interface, it is called for every node in the CDK stack that is visited as part of the synthesis process

malcyL (approved):

PR #40, opened 2022-02-08
feat: refactor cfn prefixers and add lambda prefixer

stephenrob:
Refactored the CfnResourcePrefixers to use an abstract base class and add prefixers for the following:

- CfnSecurityGroup (aws-ec2)
- CfnRole (aws-iam)
- CfnFunction (aws-lambda)

This extracts the work done on https://github.com/talis/talis-cdk-constructs/tree/sm08_add_lambda_prefixer

### DIFF
--- a/lib/aspects/resource_prefixer/resource_prefixer.ts
+++ b/lib/aspects/resource_prefixer/resource_prefixer.ts
@@ -54,7 +54,7 @@ export class ResourcePrefixer implements IAspect {
       }
     }
 
-    throw new Error("Undefined resource for resource prefixer");
+    throw new Error(`Undefined resource for resource prefixer: ${node.cfnResourceType}`);
   }
 
   private registerPrefixer(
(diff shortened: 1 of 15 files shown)


lib/aspects/resource_prefixer/resource_prefixer.ts
```diff
@@ -27,6 +33,9 @@ export class ResourcePrefixer implements IAspect {
     this.registerPrefixer(CfnTable, DynamoDbCfnTablePrefixer);
     this.registerPrefixer(CfnApi, Apigatewayv2CfnApiPrefixer);
     this.registerPrefixer(CfnStage, Apigatewayv2CfnStagePrefixer);
+    this.registerPrefixer(CfnSecurityGroup, Ec2CfnSecurityGroupPrefixer);
+    this.registerPrefixer(CfnRole, IamCfnRolePrefixer);
+    this.registerPrefixer(CfnFunction, LambdaCfnFunctionPrefixer);
```
mikerudd, line 38: IMO we should also add the node type on line 57, something along the lines of (the generic error can leave you trying to guess which prefixer is required):
 
```
"Undefined resource for resource prefixer: " + node.cfnResourceType
```

malcyL: I've added this.

malcyL (approved):

PR #44, opened 2022-03-04
fix: truncate prefixed name to 64 characters

stephenrob:
AWS expects names to be 64 characters or less, this updates the prefixer aspect to truncate names longer than 64 characters when using logical ID for the name

### DIFF
--- a/lib/aspects/resource_prefixer/cfn_resource_prefixer.ts
+++ b/lib/aspects/resource_prefixer/cfn_resource_prefixer.ts
@@ -1,5 +1,7 @@
 import { IConstruct, CfnResource, Token } from "@aws-cdk/core";
 
+const MAX_LENGTH = 64;
+
 export interface CfnResourcePrefixer {
   prefix(): void;
 }
@@ -24,18 +26,18 @@ export abstract class CfnResourcePrefixerBase implements CfnResourcePrefixer {
     name: string | undefined,
     propertyPath: string
   ): void {
+    let prefixedName;
     if (!Token.isUnresolved(name)) {
-      this.node.addPropertyOverride(
-        propertyPath,
-        `${this.resourcePrefix}${name}`
-      );
+      prefixedName = `${this.resourcePrefix}${name}`;
     } else {
       const logicalId = this.node.stack.getLogicalId(this.node);
-      this.node.addPropertyOverride(
-        propertyPath,
-        `${this.resourcePrefix}${logicalId}`
-      );
+      prefixedName = `${this.resourcePrefix}${logicalId}`;
     }
+
+    this.node.addPropertyOverride(
+      propertyPath,
+      prefixedName.substring(0, MAX_LENGTH)
+    );
   }
 
   public prefix() {
--- a/test/infra/aspects/resource_prefixer.test.ts
+++ b/test/infra/aspects/resource_prefixer.test.ts
@@ -2,6 +2,7 @@ import * as cdk from "@aws-cdk/core";
 
 import { expect as expectCDK, haveResource } from "@aws-cdk/assert";
 import { Annotations, Template } from "@aws-cdk/assertions";
+import { Role, ServicePrincipal } from "@aws-cdk/aws-iam";
 import { ResourcePrefixer } from "../../../lib";
 import { Aspects } from "@aws-cdk/core";
 import { EmptyResource } from "../../fixtures/infra/empty_resource";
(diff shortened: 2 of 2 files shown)


mikerudd (approved):

malcyL:
We have made a decision that we are not working on the code prefixer anymore.

This has been merged without being passed around for the team to review or have any input on.

malcyL:
If we do want to take this PR forward, I have the following questions:

#### 1) 

Can we expand on the description of this PR:

> AWS expects names to be 64 characters or less, this updates the prefixer aspect to truncate names longer than 64 characters when using logical ID for the name

We are making this change in response to using the CDN Site Hosting construct in the lti microservice. We have used this in multiple places before: UserUtils and Otis. For these deployments we have not had to truncate a name.

Can we detail which resource / name is too long and is having to be truncated please.

#### 2) 

I’d like to understand the specifics of the name we are truncating. What is the string before we truncate it? What is it after we have truncated it? What have we lost by truncating it? Will it **always** be safe to truncate it this way?

The test case in this PR uses a role name of `kYTwwGzerWgBAZEnEKbuUnvLzFnZhRiuDAWlmjpOZhebJYTNKOcxuJDvjwzthdiIKvjVbYmSAuIwprweKYTlOjhQtptvGPCMaFsdRuufBYBhvykpxISQbeGgDXLnFxYqZSkAjZMJchsj` which results in a final prefixed name of: `test-prefix-kYTwwGzerWgBAZEnEKbuUnvLzFnZhRiuDAWlmjpOZhebJYTNKOcx`. 

If this is a realistic example? Why do we think it is safe to truncate this? Hasn't AWS given it this string to guarantee uniqueness?   

#### 3) 

What other solutions were considered here?

Would this be a problem at all if we were following the original AWS_PREFIX pattern used elsewhere? Would this be a safer solution? Would it be quicker and safer at this point to remove the prefixer from the LTI micro service?

Can something in the CDN Site Hosting construct just be given a shorter name? My guess is that that's not a fix here, because what's different is that the prefixer is attempting to name resources which don't strictly need prefixing and are given unique names by the CDK library anyway.

mikerudd:
I have my stack setup to provide more detailed information...

> Can we detail which resource / name is too long and is having to be truncated please.

This is for a `AWS::IAM::Role`

The error message shown when deploying the stack is:

```
lti-admin/SiteCertificate/CertificateRequestorFunction/ServiceRole (ltiadminSiteCertificateCertificateRequestorFunctionServiceRoleC95F6C09) 1 validation error detected: Value 'dev-mr3-ltiadminSiteCertificateCertificateRequestorFunctionServiceRoleC95F6C09' at 'roleName' failed to satisfy constraint: Member must have length less than or equal to 64 (Service: AmazonIdentityManagement; Status Code: 400; Error Code: ValidationError; Request ID: 3cd07d97-c306-42b9-9b2f-38ccc9e5561d; Proxy: null)
```

The name causing the error in this case is:

```
dev-mr3-ltiadminSiteCertificateCertificateRequestorFunctionServiceRoleC95F6C09
```
length: 78 characters

stephenrob:
> #### 1)
> Can we expand on the description of this PR:
> 
> > AWS expects names to be 64 characters or less, this updates the prefixer aspect to truncate names longer than 64 characters when using logical ID for the name
> 
> We are making this change in response to using the CDN Site Hosting construct in the lti microservice. We have used this in multiple places before: UserUtils and Otis. For these deployments we have not had to truncate a name.
> 
> Can we detail which resource / name is too long and is having to be truncated please.


Currently the names are generated as follows by the CDK

Screenshot 2022-03-07 at 09 36 23
Screenshot 2022-03-07 at 09 35 49

> #### 3)
> What other solutions were considered here?
> 
> Would this be a problem at all if we were following the original AWS_PREFIX pattern used elsewhere? Would that be a safer solution? Would it be quicker and safer at this point to remove the prefixer from the LTI micro service?
>

We wouldn't prefix any of the roles or resources created by the construct and other constructs used within this, going to unprefixed auto generated names
 
>
> Can something in the CDN Site Hosting construct just be given a shorter name? My guess is that that's not a fix here, because what's different is that the prefixer is attempting to name resources which don't strictly need prefixing and are given unique names by the CDK library anyway.

If we don't want to prefix names that are generated by the CDK library then we shouldn't statically name any resources within the stack unless absolutely necessary i.e. bucket names or dns names and rely on the auto generated names to guarantee uniqueness. Although we aren't sure how this will work bringing up multiple stacks in the same account.

malcyL:
To clarify then: we are truncating:

```
dev-mr3-ltiadminSiteCertificateCertificateRequestorFunctionServiceRoleC95F6C09
```

to

```
dev-mr3-ltiadminSiteCertificateCertificateRequestorFunctionServi
```

I'm not sure this is the correct thing to do.

PR #47, opened 2022-05-09
feat: plat-5537 authenticated api updates from from smalti

malcyL:
- https://github.com/talis/platform/issues/5537

This PR pulls in modifications to the Authenticated API Construct made in the lti microservice project. That project did not start by using the construct form this library, instead recreating it in it's own repo. It followed the design of the AuthenticatedAPI construct from this repo, but made improvements to it - hence this PR to retro fit those improvements before it is used elsewhere.

The improvements are:

- The AuthenticatedAPI no longer creates the lambda's inside the construct. Instead the lambda are created outside and passed in. This allows for permissions to be added to the lambda in the standard cdk was rather than having to pass policy statements into the api construct.
- The AuthenticatedAPI now creates a domain name
- An extra cloudwatch alarm has been added to report on lmabda errors.
- We export the name of the API to the console for ease of use
- The Authorizer function is now typescript


package.json
```diff
@@ -15,7 +15,7 @@
     "integration-test": "npm run build && npm run jest-integration-test",
     "jest-integration-test": "jest test/integration/",
     "lint:prettier": "prettier --check .",
-    "lint:eslint": "eslint . --ext .js,.ts",
+    "lint:eslint": "eslint . --ext .ts",
```
stephenrob, line 18: Removing `js` from the linter is because we've moved the authorizer to typescript ?

stephenrob (approved):
Looks good - Doesn't seem to be any issues when its been moved from SMALTI

package.json
malcyL: Yes. This was there to lint the original `authorizer.js` file. We don't want to lint it anymore, now it's the output of compiling the `.ts` file.

PR #52, opened 2022-05-18
fix!: deploy SPA websiteIndexDocument last

camillef:
BREAKING CHANGE: Routed SPAs must specify a websiteIndexDocumentAsset that points to the source of the websiteIndexDocument asset.

For routed SPAs, deploy and invalidate websiteIndexDocument only after all other assets have been deployed, to ensure deployment does not cause downtime.

### DIFF
--- a/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
+++ b/lib/cdn-site-hosting/cdn-site-hosting-construct.ts
@@ -31,10 +31,6 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
   ) {
     super(scope, id);
 
-    if (props.websiteIndexDocument.startsWith("/")) {
-      throw Error("leading slashes are not allowed in websiteIndexDocument");
-    }
-
     validateProps(props);
 
     const siteDomain = getSiteDomain(props);
@@ -109,7 +105,7 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
         props.sourcesWithDeploymentOptions.length === 1;
 
       // multiple sources with granular cache and invalidation control
-      props.sourcesWithDeploymentOptions.forEach(
+      const deployments = props.sourcesWithDeploymentOptions.map(
         (
           { name, sources, distributionPathsToInvalidate, cacheControl },
           index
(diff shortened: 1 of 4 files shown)


lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -31,10 +31,6 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
   ) {
     super(scope, id);
 
-    if (props.websiteIndexDocument.startsWith("/")) {
-      throw Error("leading slashes are not allowed in websiteIndexDocument");
-    }
```
camillef, line 36: Broken window: moved into `validateProps` function below

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -120,20 +116,47 @@ export class CdnSiteHostingConstruct extends cdk.Construct {
...
+            destinationBucket: this.s3Bucket,
+            distribution: this.cloudfrontWebDistribution,
+            distributionPaths: [`/${props.websiteIndexDocument}`],
+          }
+        );
+        deployments.forEach((deployment) => {
+          websiteIndexDocumentDeployment.node.addDependency(deployment);
+        });
+      }
     } else if (props.sources) {
```
camillef, line 160: Question: do we need to do the same if `props.sources` is provided? What are the respective use cases for these props?

lib/cdn-site-hosting/cdn-site-hosting-props.ts
```diff
@@ -17,4 +17,5 @@ export interface CommonCdnSiteHostingProps {
   sourcesWithDeploymentOptions?: SourcesWithDeploymentOptions[];
   websiteErrorDocument?: string;
```
camillef, line 18: Question: do we need any custom treatment for when a separate `websiteErrorDocument` is provided for an SPA? Not sure of the use case for this either

camillef:
Closing in favour of a simpler approach: https://github.com/talis/talis-cdk-constructs/pull/53

PR #54, opened 2022-06-14
chore: re-enable commit linting

kiyanwang:
- This relates to talis/platform#5578

This P/R achieves the following
- [x] re-introduces husky locally for those developers who are happy to use it to prevent them from violating conventional commits by using `commitlint` as a git hook.
- [x] introduces a new workflow step `commit-lint-branch` which will fail the build (so you won't be able to merge) but does not prevent a developer from running tests on their branch.
  - [x] for this to work correctly we need to work out when the branch was created and use that commit as the starting point or `--from`. The first actual commit on a branch cannot be used because it checks every commit *after* it but not including.
  - [x] this is the difference btw `git merge-base main <branch>` vs ` git log main..<branch> --oneline | tail -1 | awk '{ print $1 }'`
- [x] Finally the `commit-lint-branch` step will not run against `main` when your branch is merged. This is because there are numerous commits on main that already violate conventional commits.

## Failing build, does not prevent dev running tests on branch

image

## output from job (deliberately set to verbose)

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,12 @@ workflows:
   build:
     jobs:
       - fan_out
-      - commit-lint-branch
+      - commit-lint-branch:
+          filters:
+            branches:
+              ignore:
+                - main
+                - /^dependabot-.*$/
       - be_kind_to_your_colleagues:
           type: approval
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,19 @@ jobs:
       - run:
           command: echo "starting build"
       - save_root_build_number
-
+  commit-lint-branch:
+    executor: default_executor
+    steps:
+      - add_ssh_keys:
+          fingerprints:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -41,7 +41,22 @@ jobs:
...
+      - add_ssh_keys:
+          fingerprints:
+            - "ee:b4:00:e6:23:5b:55:bb:fd:07:bc:73:9e:f7:89:9c" # shared.aspire@talis.com 'talis-cdk-constructs Deploy Key'
+      - checkout
+      - node/install-packages
+      - run:
+          name: Lint all commits on the branch
+          command: |
+            FIRST_COMMIT=$(git merge-base main $CIRCLE_BRANCH)
+            LAST_COMMIT=$(git rev-parse HEAD)
```
mrliptontea, line 56: Is that how we found the commit range previously?

mrliptontea (approved):
Looks broadly OK. I'm just a bit unsure about the logic for finding commit range.

.circleci/config.yml
kiyanwang: no, AFAIK we were never checking a range, only the last commit.

ShaunParsons (approved):

PR #55, opened 2022-06-22
feat: plat-5642 add support for none lambda routes to authenticatedapi

malcyL:
- https://github.com/talis/platform/issues/5642

Several of our serverless apps host API documentation auto generated from the app. This requires routing to static files rather than routing to a lambda.

We still want to maintain the ability to add authorisation, monitoring, enforce specific settings etc around routes, so we do not want to expose the api outside of our construct.

This PR adds the ability to set a simple url route to direct to static content set up elsewhere. This will allow an applications api gateway to route to open api documentation generated by the project.

This PR renames `routes` to `lambdaRoutes` in the constructs props and also adds `urlProps`. Additionally, we have decided to create another way to add these routes - with a new method `addUrlRoute`. We envisage making it possible to add lambda routes in this way in the future too - but that's not done here.

This PR also refactors the `paths` in the lambda route properties to be `path` - singular rather than an array. Adding the url route I added this as a singular path, which pointed out that this should really be a singular path too. Everywhere we have used it - it is a single path in an array, which again points out that it was wrong. 

This is an example in `smalti` where we add routes to redirect to the documentation. Adding this here will mean that `smalti` can use this construct:

https://github.com/talis/smalti-server/blob/main/packages/infra/src/stacks/app-stack.ts#L510

```typescript
    api.gateway.addRoutes({
      path: '/1/docs',
      methods: [apigateway2.HttpMethod.GET],
      integration: new apigateway2Integrations.HttpUrlIntegration(
        'lti-api-documentation',
        `${ltiBucket.bucketWebsiteUrl}/api-documentation/index.html`,
        {
          method: apigateway2.HttpMethod.GET,
        }
      ),
    });

    api.gateway.addRoutes({
      path: '/1/open-api',
      methods: [apigateway2.HttpMethod.GET],
      integration: new apigateway2Integrations.HttpUrlIntegration(
        'lti-api-documentation',
        `${ltiBucket.bucketWebsiteUrl}/api-documentation/open-api.yml`,
        {
          method: apigateway2.HttpMethod.GET,
        }
      ),
    });

```

### DIFF
--- a/lib/authenticated-api/authenticated-api.ts
+++ b/lib/authenticated-api/authenticated-api.ts
@@ -20,9 +20,10 @@ const DEFAULT_LAMBDA_DURATION_THRESHOLD = cdk.Duration.minutes(1);
 export class AuthenticatedApi extends cdk.Construct {
   readonly apiId: string;
   readonly httpApiId: string;
-  readonly httpApi: apigatewayv2.HttpApi;
-  readonly authorizer: apigatewayv2.IHttpRouteAuthorizer;
-  readonly alarmAction: IAlarmAction;
+
+  private httpApi: apigatewayv2.HttpApi;
+  private authorizer: apigatewayv2.IHttpRouteAuthorizer;
+  private alarmAction: IAlarmAction;
 
   constructor(scope: cdk.Construct, id: string, props: AuthenticatedApiProps) {
     super(scope, id);


examples/simple-authenticated-api/lib/simple-authenticated-api-stack.ts
```diff
@@ -111,23 +112,47 @@ export class SimpleAuthenticatedApiStack extends cdk.Stack {
           oauth_route: "/oauth/tokens/",
         },
 
-        routes: [
+        lambdaRoutes: [
           {
             name: "route1",
-            paths: ["/1/route1"],
+            path: "/1/route1",
```
malcyL, line 118: Changing this from paths -> path to make it match the new url routes.

The url routes can be added by a method as well as props - which highlighted that this should always have been a singular path. We have only ever used a single path here - also telling us that it was wrong.

This is a breaking change though - so  I will put it in the release notes.

lib/authenticated-api/authenticated-api-props.ts
```diff
@@ -3,13 +3,15 @@ import * as ec2 from "@aws-cdk/aws-ec2";
...
+import { RouteUrlProps } from "./route-url-props";
 
 export interface AuthenticatedApiProps {
   prefix: string;
   name: string;
   description: string;
   stageName: string;
-  routes: Array<RouteLambdaProps>;
+  lambdaRoutes?: Array<RouteLambdaProps>;
+  urlRoutes?: Array<RouteUrlProps>;
```
malcyL, line 14: routes -> lambdaRoutes - another breaking change - but this better distinguishes it from urlRoutes.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -50,29 +56,29 @@ export class AuthenticatedApi extends cdk.Construct {
       }),
     };
 
-    const httpApi = new apigatewayv2.HttpApi(
+    this.httpApi = new apigatewayv2.HttpApi(
```
malcyL, line 59: `httpApi` is now a member so it can be accessed in the new addUrlRoute method.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -50,29 +56,29 @@ export class AuthenticatedApi extends cdk.Construct {
...
+    this.alarmAction = new cloudwatchActions.SnsAction(props.alarmTopic);
 
     // Routes may contain required scopes. These scopes need to be in the config
     // of the authorization lambda. Create this config ahead of creating the authorization lambda
     const scopeConfig: { [k: string]: string } = {};
-    for (const routeProps of props.routes) {
-      if (routeProps.requiredScope) {
-        for (const path of routeProps.paths) {
-          scopeConfig[`^${path}$`] = routeProps.requiredScope;
+    if (props.lambdaRoutes) {
```
malcyL, line 78: I made `lambdaRoutes`  optional - as there could be just `urlRoutes`.

examples/simple-authenticated-api/README.md
```diff
@@ -7,6 +7,7 @@ The AuthenticatedApi:
 - creates an ApiGateway (v2)
 - creates an authorisation lambda which authenticates tokens against persona
 - creates any number of routes, each with a lambda to handle requests.
+- creates any number of routes redirecting to a given url.
```
kiyanwang, line 10:
```suggestion
- creates any number of routes redirecting to a given url(s).
```

examples/simple-authenticated-api/lib/simple-authenticated-api-stack.ts
```diff
@@ -111,23 +112,47 @@ export class SimpleAuthenticatedApiStack extends cdk.Stack {
...
+        publicReadAccess: true,
+        removalPolicy: cdk.RemovalPolicy.DESTROY,
+        websiteIndexDocument: "index.html",
+      }
+    );
+
+    // Url Routes can be added in the initial props of the api construct, but they can also be
+    // added using the following method
+    api.addUrlRoute({
+      name: "simple authenticated api docs",
```
kiyanwang, line 152: to be clear, this construct isn't adding authentication to the route for the documentation.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -128,92 +134,98 @@ export class AuthenticatedApi extends cdk.Construct {
...
-          comparisonOperator:
-            cloudwatch.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
-          // Set treatMissingData to IGNORE
-          // Stops alarms with minimal data having false alarms when they transition to this state
-          treatMissingData: cloudwatch.TreatMissingData.IGNORE,
-        }
-      );
-      durationAlarm.addAlarmAction(alarmAction);
-      durationAlarm.addOkAction(alarmAction);
+        // Add an alarm on the duration of the lambda dealing with the HTTP Request
```
kiyanwang, line 167: is this indentation correct, it looks like you might a level too deep now.

kiyanwang (approved):
This looks reasonable. I still get confused re: the naming of the authenticated api, maybe that's something we should revisit later.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -10,13 +11,18 @@ import * as lambdaNodeJs from "@aws-cdk/aws-lambda-nodejs";
...
+import { RouteUrlProps } from "./route-url-props";
+import { IAlarmAction } from "@aws-cdk/aws-cloudwatch";
 
 const DEFAULT_API_LATENCY_THRESHOLD = cdk.Duration.minutes(1);
 const DEFAULT_LAMBDA_DURATION_THRESHOLD = cdk.Duration.minutes(1);
 
 export class AuthenticatedApi extends cdk.Construct {
   readonly apiId: string;
   readonly httpApiId: string;
+  readonly httpApi: apigatewayv2.HttpApi;
```
stephenrob, line 23: By adding this as `readonly` it is now accessible outside of the class i.e.

```
const myApi = new AuthenticatedApi(...)

myApi.httpApi.whateverMethodIWantToCall(...)
```

As we've discussed we don't want to expose the underlying httpApi outside of the class so this should defined as private readonly if you don't want to change it once set or just private if you do

```
private readonly httpApi: apigatewayv2.HttpApi
```

PR #56, opened 2022-07-07
feat: plat-5577 api construct exposes the domain name

malcyL:
- https://github.com/talis/platform/issues/5577

This PR makes the `domainName` object a read only attribute from the API construct.

This is needed for Smalti which is using the regional domain name and the regional hosted zone id.


examples/simple-authenticated-api/lib/simple-authenticated-api-stack.ts
```diff
@@ -154,5 +154,10 @@ export class SimpleAuthenticatedApiStack extends cdk.Stack {
       path: "/api-documentation",
       method: apigatewayv2.HttpMethod.GET,
     });
+
+    console.log(`Regional domain name: ${api.domainName.regionalDomainName}`);
```
stephenrob, line 158: Could register these as a CFN Output from the stack but this is fine to show they are available

PR #57, opened 2022-07-12
fix: correctly identify scopes for paths containing arguments

malcyL:
- https://github.com/talis/platform/issues/5577

This PR fixes a bug in the AuthenticatedAPI construct that we have introduced by bringing in the Typescript authorizer.ts file from Smalti.

## Problem

The regex used to identify paths requiring scopes is not matching against paths which conatin an argument.

i.e. it's checking that `^/1/registration/{id}$` matches against `/1/registration/1234` and failing because it's not interpreting `{id}` correctly.

## Solution

Having tried a coupe of libraries (https://github.com/pillarjs/path-to-regexp and https://github.com/mswjs/node-match-path) which did not work because they expected express syntax `:arg` rather than aws syntax `{arg}`, I discussed this with Nad who suggested we implement our own matcher, which is reasonably trivial.

This PR therefore refactors the part of the authorizer which is matching paths to their required scopes to use a new matching function which correctly matches paths containing arguments.

We are using the simple example to deploy and run integration tests against. Two new routes are therefore added to the simple example and integration tests are added to ensure they work as expected. This feels like we are extending the simple example to include test cases, and perhaps the tests will need their own deployment in the future to keep the simple example simple. But for now - I have added two routes to it.

## Why did this work in smalti?

We're making this refactor so we can pull this into smalti - but how did this work in that project?

smalti's copy of the construct exposes the internals of the authorizer - specifically the entire path/scope config implementation. This allows smalti to set this config up to work. We do not want to expose these internals out of the construct, so that when the implementation changes, the API does not need to. Therefore, the construct sticks with the approach of specifying a scope per method, not setting the scope config up for the internal implementation.


src/lambda/api/authorizer.ts
```diff
@@ -241,6 +241,31 @@ class PersonaAuthorizer {
 
     return this.personaClient;
   }
+
+  pathMatch(pathDefination: string, path: string): boolean {
+    const pathDefinationParts = pathDefination.split("/");
+    const pathParts = pathDefination.split("/");
```
mrliptontea, line 247: Looks like a mistake:

```suggestion
    const pathParts = path.split("/");
```

Which makes me think we're missing some test coverage?

ShaunParsons (approved):

src/lambda/api/authorizer.ts
malcyL: Done

PR #59, opened 2022-07-14
feat: ele-4721 update cdk to latest v1.163.1

Crevitus:
Upgrade the AWS CDK dependency from 1.143.0 to 1.163.1. This brings in the latest changes from the v1 branch of the CDK which provides the [sslsupportmethod](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-cloudfront.Distribution.html#sslsupportmethod) property in the `Distribution` class.

This is required for switching over the `cdn-site-hosting-construct` to use `cloudfront.Distribution` which is required to add security headers for Otis.

Also fixes @types/prettier to <2.6.0 to resolve a bunch of build errors that started happening after the upgrade - see comment and discussion here - https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/60310#discussioncomment-2811465

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@types/lodash": "4.14.178",
     "@types/node": "10.17.27",
     "@types/uuid": "^8.3.1",
+    "@types/prettier": "<2.6.0",
     "@typescript-eslint/eslint-plugin": "^4.28.3",
     "@typescript-eslint/parser": "^4.28.3",
     "aws-sdk": "^2.1040.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2360,9 +2360,9 @@
       "dev": true
     },
     "@types/prettier": {
-      "version": "2.6.3",
-      "resolved": "https://registry.npmjs.org/@types/prettier/-/prettier-2.6.3.tgz",
-      "integrity": "sha512-ymZk3LEC/fsut+/Q5qejp6R9O1rMxz3XaRHDV6kX8MrGAhOSPqVARbDi+EZvInBpw+BnCX3TD240byVkOfQsHg==",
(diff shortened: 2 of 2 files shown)


camillef (commented):
Hate to say this, but 1.163.2 has been released

Crevitus:
> Hate to say this, but 1.163.2 has been released

They're bumping it all the time and this has what I need, it will take a while to regen all the package-locks, I am okay with leaving it as is.

package.json
```diff
@@ -47,31 +48,32 @@
...
-    "@aws-cdk/aws-sqs": "1.143.0",
-    "@aws-cdk/core": "1.143.0",
+    "@aws-cdk/aws-apigatewayv2": "1.163.1",
+    "@aws-cdk/aws-apigatewayv2-authorizers": "1.163.1",
+    "@aws-cdk/aws-apigatewayv2-integrations": "1.163.1",
+    "@aws-cdk/aws-certificatemanager": "1.163.1",
+    "@aws-cdk/aws-cloudfront": "1.163.1",
+    "@aws-cdk/aws-cloudwatch": "1.163.1",
+    "@aws-cdk/aws-cloudwatch-actions": "1.163.1",
+    "@aws-cdk/aws-cloudfront-origins": "1.163.1",
```
camillef, line 61: So this is the new dependency?

camillef (approved):

package.json
Crevitus: Oh this is a new one to support the change over to the new class - it shouldn't technically be in this PR as its not used yet...

Crevitus: Leaving this for now as its not much harm being there.

PR #61, opened 2022-08-24
feat: add Lambda Ephemeral storage to LambdaWorker props

ShaunParsons:
Add support for `ephemeralStorageSize` in the `LambdaWorker` construct.


lib/lambda-worker/lambda-worker-props.ts
```diff
@@ -20,6 +20,7 @@ export interface LambdaWorkerProps {
     entry?: string;
     enableQueue?: boolean;
     environment?: { [key: string]: string };
+    ephemeralStorageSize?: cdk.Size;
```
stephenrob, line 23: Do we need to add any validation that it is less than 10240 or does the underlying L2 construct handle this

stephenrob (approved):
Looks good - just a non blocking question on validation of the prop

lib/lambda-worker/lambda-worker-props.ts
ShaunParsons: Yup: https://github.com/aws/aws-cdk/blob/1ef0af8dfe6dfb9653338ea9daa6a60d8cbd8ccb/packages/%40aws-cdk/aws-lambda/lib/function.ts#L766-L769

malcyL (approved):

PR #67, opened 2023-02-28
feat!: plat-6337 cdk stack adds tfs tags in construct

juan-casimiro:
- https://github.com/talis/platform/issues/6337

Note: This is a breaking change, users of the _TalisCdkStack_ will need to add new non-optional properties to the _TalisCdkStackProps_.

On this PR:
- **TalisCdkStackProps**: Added new **non-optional** properties _release_ and _app_.
- **TalisCdkStack** adds the following tags to all taggable resources created within the construct:
    - _tfs-app_
    - _tfs-environment_
    - _tfs-release_
    - _tfs-region_
    - _tfs-service_
- Created new _example_ SimpleTalisCdkStack ([_/examples/simple-talis-cdk-stack.ts_](https://github.com/talis/talis-cdk-constructs/blob/PLAT-6337-cdk/examples/simple-talis-cdk-stack/bin/simple-talis-cdk-stack.ts))
- readme files updated.

This screenshot shows the tags set in the SimpleTalisCdkStack example:

image


malcyL:
This looks good @juan-casimiro - but as we worked on this together, I'll leave it for someone else to review/thumb as well as me.

Might be worth adding a screen shot into the PR description showing the tags on the stack?

lib/talis-cdk-stack/talis-cdk-stack.ts
```diff
@@ -11,6 +13,23 @@ export class TalisCdkStack extends Stack {
...
+    cdk.Tags.of(scope).add("tfs-environment", props.deploymentEnvironment); // e.g. production staging development
+    cdk.Tags.of(scope).add("tfs-release", props.release); // e.g. 8561-105814f
+
+    // props.env comes from aws cdk core StackProps. It's optional
+    // so that an environment agnostic stack can be created.
+    // If we create an environment agnostic stack - it can not contain
+    // region or service.
+    // However, when we deploy into our environments for real - props.env will be set
+    if (props.env?.region) {
+      cdk.Tags.of(scope).add("tfs-region", props.env.region);
```
mrliptontea, line 27: for `tfs-region` we meant the short region:
https://github.com/talis/talis-cdk-constructs/blob/902907814f31b61fb34fa7e24d845f87441a7829/lib/talis-cdk-stack/talis-region.ts#L7-L11

lib/talis-cdk-stack/talis-cdk-stack.ts
```diff
@@ -11,6 +13,23 @@ export class TalisCdkStack extends Stack {
...
+      cdk.Tags.of(scope).add(
+        "tfs-service",
+        `${props.app}-${props.env.region.split("-")[0]}`
+      );
+    }
```
mrliptontea, line 32: This isn't formally documented anywhere, but generally, the convention is that in production we simply use `<app>-<region>` but use `<app>-<environment>-<region>` for other environments, e.g. `depot-staging-eu`, `otis-build-eu`, etc.

This has been somewhat applied to URLs we give our services, for example:
- https://files.talis.com/ - Depot server production
- https://files.ca.talis.com/ - Depot server production Canada
- https://staging-files.talis.com/ - Depot server staging

lib/talis-cdk-stack/talis-region.ts
```diff
@@ -9,3 +9,20 @@ export enum TalisShortRegion {
...
+  talisRegionName: string
+): TalisShortRegion | undefined {
+  const talisRegion = Object.values(TalisRegion).find(
+    (talisRegion) => talisRegion === talisRegionName
+  );
+  if (!talisRegion) {
+    return undefined;
+  }
+  return talisRegionToShortRegionMap.get(talisRegion);
+}
```
mrliptontea, line 28: While your approach is valid, it's a bit verbose.
In JavaScript/TypeScript we could write it shorter:

```suggestion
const talisRegionToShortRegionMap = {
  [TalisRegion.CANADA]: TalisShortRegion.CANADA,
  [TalisRegion.EU]: TalisShortRegion.EU,
  [TalisRegion.LOCAL]: TalisShortRegion.LOCAL,
};

export function getTalisShortRegionFromTalisRegion(
  talisRegion: TalisRegion
): TalisShortRegion | undefined {
  return talisRegionToShortRegionMap[talisRegion] ?? undefined;
}
```

There's nothing wrong with using Maps over plain Objects, but Maps are mostly useful when they are filled dynamically and/or when keys are not primitive values. For static mapping, object literal syntax is easier to use and read.

mrliptontea (approved):

lib/talis-cdk-stack/talis-cdk-stack.ts
juan-casimiro: agreed and updated

PR #68, opened 2023-03-01
feat: commit used for triggering a release pipeline

juan-casimiro:
BREAKING CHANGES: added new non-optional TalisCdkStackProps-> app, release


malcyL:
This is my understanding of what we are doing here.

#### What Happened

The previous PR had followed the instructions here:

https://www.conventionalcommits.org/en/v1.0.0/

image

to use `feat!` to indicate a breaking change.

What actually happened was:

https://app.circleci.com/pipelines/github/talis/talis-cdk-constructs/512/workflows/760177d2-5593-433f-9c37-1a46c24df975/jobs/1653

image

```
[3:03:34 PM] [semantic-release] [@semantic-release/commit-analyzer] › ℹ  Analyzing commit: feat!: plat-6337 cdk stack adds tfs tags in construct (#67)
[3:03:34 PM] [semantic-release] [@semantic-release/commit-analyzer] › ℹ  The commit should not trigger a release
[3:03:34 PM] [semantic-release] [@semantic-release/commit-analyzer] › ℹ  Analysis of 1 commits complete: no release
[3:03:34 PM] [semantic-release] › ✔  Completed step "analyzeCommits" of plugin "@semantic-release/commit-analyzer"
[3:03:34 PM] [semantic-release] › ℹ  There are no relevant changes, so no new version is released.
```

So - instead of following Convention Commits V1.0.0 - the library took the commit message for a **breaking** change as a commit message for a commit **not needing a release**!

#### What We Want To Achieve In This PR

What we want this PR to do is:
- force a release of the library
- make the semantic versioning trigger a breaking change release

Looking at this PR it has one commit:

image

Because this PR has only one commit message, we expect a GitHub PR process to use the commit message, not the PR title. We are hoping that this commit message will trigger a breaking change release:

```
feat: commit used for triggering a release pipeline

BREAKING CHANGES: added new non-optional TalisCdkStackProps-> app, release
```

@juan-casimiro were you able to add this commit message in the normal way and it pass branch commit message linting?
I am wondering if it was accepted?

If it wasn't accepted - can we add to this PR the details of what we had to do to get it accepted please.

mrliptontea (changes requested):
Instead of these changes we should update the semantic-release config to actually use conventional commits, see commit https://github.com/talis/talis-cdk8s-constructs/commit/86064d9404e326723aa7872eaed47d361f4bfd9e.

This is because semantic-release package by default uses a convention similar to [conventionalcommits](https://www.conventionalcommits.org/), but not exactly it. By setting the config we will make it use the same rules implementation as commitlint-cli, so linting will be accurate.

Then, since the previous merge didn't trigger a release, after making this change, it will.

README.md
```diff
@@ -73,3 +73,34 @@ To run the integration tests locally:
...
+Be aware of GitHub PRs feature _reviewers code suggestions_: accepting & commiting suggestions using the GitHub GUI will generate commits with coauthor information in the comment that will break the _commitlint_ rules.
+
+### _semantic-release_
+
+_Breaking Change_: _semantic-release 17.4.4_ doesn't support _feat!:_ as a _breaking change_ commit type, and it won't recognize it as a _feat_ type either, so as a result it won't publish a new release.
+
+Use the commit comment footer _BREAKING CHANGE:..._ instead. This will trigger a new _Breaking Release_.
+
+To check if your branch follows the _semantic-release_ rules, run the command
+`npx semantic-release -b add_release_label_to_the_namespace --dry-run --branches main`
```
mrliptontea, line 104: The `-b` at the start _is_ the `--branches` flag, but `add_release_label_to_the_namespace` was the name of my branch on talis-cdk8s-constructs... For the check to work you push changes to your branch and then set the flag with your branch name, so that you check before you merge to `main` (which is the default when `--branches` is not set).

:neutral_face:

juan-casimiro: I see! thanks Grz

mrliptontea (approved):

PR #71, opened 2023-04-24
chore: update dependencies

mrliptontea:
* Update all dependencies to their latest versions
* Should resolve the following in projects:

    > 21138	(aws-ec2): VPCs with SubnetType.PRIVATE cause resource destruction on versions >=v1.162.0, <=1.163.2
    > 
    > 	Overview: On upgrading to versions >=v1.162.0 & <=1.163.2 and using
    > 	          SubnetType.PRIVATE in a VPC subnetConfiguration, some of the
    > 	          VPC resources will be destroyed.
    > 
    > 	Affected versions: framework: >=1.162.0 <=1.163.2
    > 
    > 	More information at: https://github.com/aws/aws-cdk/issues/21138
    >


test/integration/authenticated-api/authenticated-api.test.ts
```diff
@@ -113,7 +113,7 @@ describe("AuthenticatedApi", () => {
       await axiosNoAuthInstance.get("route1");
       throw Error("Expected a 401 response");
     } catch (err) {
-      expect(err.message).toBe("Request failed with status code 401");
+      expect(err).toMatch("Request failed with status code 401");
```
mrliptontea, line 116: This works around the issue that `err` is `any`. A stringified Error will take format: `Error: {message}` so this will match.

lib/lambda-worker/lambda-worker.ts
```diff
@@ -296,7 +296,7 @@ export class LambdaWorker extends cdk.Construct {
 
     return new lambda.DockerImageFunction(this, props.name, {
       code: lambda.DockerImageCode.fromEcr(ecrRepository, {
-        tag: imageTag,
+        tagOrDigest: imageTag,
```
mrliptontea, line 299: There was a deprecation warning about `tag` property - this is the replacement.

.circleci/config.yml
```diff
@@ -106,13 +106,13 @@ jobs:
...
-          command: npm run unit-test
+          command: npm run jest-unit-test -- --ci --maxWorkers=2
       - run:
           name: Run infra tests
-          command: npm run infra-test
+          command: npm run jest-infra-test -- --ci --maxWorkers=2
       - run:
           name: Run integration tests
-          command: npm run integration-test
+          command: npm run jest-integration-test -- --ci --maxWorkers=2
```
mrliptontea, line 115: Changed it to call jest directly so we don't run `build` with each. Also, saw weird issues with jest workers being killed in CircleCI, so reduced number of maxWorkers.

src/lambda/api/PersonaAuthorizer.d.ts
```diff
@@ -1,5 +1,5 @@
 import { PersonaClient } from "talis-node";
-declare type ParsedArn = {
+type ParsedArn = {
```
mrliptontea, line 2: This change happened automatically.

.circleci/config.yml
kiyanwang: How many workers was it running by default?

kiyanwang (approved):
That all looks reasonable. Nice one.

PR #72, opened 2023-04-28
feat: enable access logs on AuthenticatedApi

mrliptontea:
* talis/platform#6429

### DIFF
--- a/lib/authenticated-api/authenticated-api.ts
+++ b/lib/authenticated-api/authenticated-api.ts
@@ -2,9 +2,11 @@ import * as acm from "@aws-cdk/aws-certificatemanager";
 import * as apigatewayv2 from "@aws-cdk/aws-apigatewayv2";
 import * as apigateway2Integrations from "@aws-cdk/aws-apigatewayv2-integrations";
 import * as authorizers from "@aws-cdk/aws-apigatewayv2-authorizers";
+import * as awslogs from "@aws-cdk/aws-logs";
 import * as cdk from "@aws-cdk/core";
 import * as cloudwatch from "@aws-cdk/aws-cloudwatch";
 import * as cloudwatchActions from "@aws-cdk/aws-cloudwatch-actions";
+import * as iam from "@aws-cdk/aws-iam";
 import * as integrations from "@aws-cdk/aws-apigatewayv2-integrations";
 import * as lambda from "@aws-cdk/aws-lambda";
 import * as lambdaNodeJs from "@aws-cdk/aws-lambda-nodejs";
@@ -46,8 +48,9 @@ export class AuthenticatedApi extends cdk.Construct {
         props.certificateArn
       ),
     });
+    const apiName = `${props.prefix}${props.name}`;
     const apiGatewayProps: apigatewayv2.HttpApiProps = {
-      apiName: `${props.prefix}${props.name}`,
+      apiName: apiName,
       defaultDomainMapping: { domainName: this.domainName },
       ...(props.corsDomain && {
         corsPreflight: {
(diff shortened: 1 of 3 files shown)


lib/authenticated-api/authenticated-api.ts
```diff
@@ -59,17 +62,56 @@ export class AuthenticatedApi extends cdk.Construct {
...
 
     this.apiId = this.httpApi.apiId;
     this.httpApiId = this.httpApi.httpApiId;
 
+    if (props.logging?.enabled) {
+      const logGroup = new awslogs.LogGroup(this, "apiGatewayLogs", {
+        // CloudWatch Logs automatically enables the /aws/vendedlogs/* prefix
+        // in the resource policy, so we use it to avoid adding each log
+        // group to the policy, which has a limit of 5120 characters.
+        logGroupName: `/aws/vendedlogs/${apiName}-accessLog`,
```
mrliptontea, line 75: When you enable logging on APIGateway, AWS automatically tries to put your resource's ARN into a resource policy called "AWSLogDeliveryWrite20150319". Trying to deploy a new stack, I got this error:

> Cannot enable logging. Policy document length breaking Cloudwatch Logs Constraints, either < 1 or > 5120

The [documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AWS-logs-and-resource-policy.html#AWS-logs-infrastructure-CWL) explains this:

> ### Log group resource policy size limit considerations
>
> These services must list each log group that they're sending logs to in the resource policy, and CloudWatch Logs resource policies are limited to 5120 characters. A service that sends logs to a large number of log groups may run into this limit.
>
> To mitigate this, CloudWatch Logs monitors the size of resource policies used by the service that is sending logs, and when it detects that a policy approaches the size limit of 5120 characters, CloudWatch Logs automatically enables /aws/vendedlogs/* in the resource policy for that service. You can then start using log groups with names that start with /aws/vendedlogs/ as the destinations for logs from these services.

So in order to avoid having issues the policy again, it's best to use the `/aws/vendedlogs/` prefix for the log group name.

Our policy was full of build and development resources, so I edited it to remove them.
Resource policies cannot be edited in AWS Console, but can be seen in the CLI with:

```sh
aws logs describe-resource-policies

# to see policy JSON:
aws logs describe-resource-policies --query 'resourcePolicies[?policyName==`AWSLogDeliveryWrite20150319`].policyDocument' --output json | jq '.[0] | fromjson'

# to update the policy:
aws logs put-resource-policy --policy-name AWSLogDeliveryWrite20150319 --policy-document "<policy JSON>"
```

There is also an option to manage that policy with Terraform via [`api_gateway_account`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_account) resource.

lib/authenticated-api/authenticated-api.ts
```diff
@@ -19,11 +21,11 @@ const DEFAULT_API_LATENCY_THRESHOLD = cdk.Duration.minutes(1);
 const DEFAULT_LAMBDA_DURATION_THRESHOLD = cdk.Duration.minutes(1);
 
 export class AuthenticatedApi extends cdk.Construct {
+  readonly httpApi: apigatewayv2.HttpApi;
   readonly apiId: string;
   readonly httpApiId: string;
   readonly domainName: apigatewayv2.DomainName;
 
-  private httpApi: apigatewayv2.HttpApi;
```
mrliptontea, line 26: I exposed the ApiGateway here so we could something like we do here in the application code consuming this construct. However, it's not required for the purpose of this so I don't know if it's needed?

malcyL: When we discussed this in the past - the original idea was to keep things in the construct. Shaun thought this would allow us to enforce things, and exposing them out of the construct would allow people to set things we didn't want them too. E.g. minimum memory size on a lambda.

We changed the Lambda worker to create a Lambda Worker Function which could be accessed outside, for example to allow it to be used for permission, but still wrapped the CDK class.

Exposing this here does seem to break the original idea of encapsulation, but the way the CDK works, maybe that encapsulation isn't going to work.

So I'm in two minds as to whether I think this is a good idea. 

Falling back to "it's easier to make things public than to make things private once they have been public" - if it's not required - is it worth not making it public for now?

lib/authenticated-api/authenticated-api.ts
malcyL: Was this the reason we backed this out after it was added in the first time?

malcyL (approved):

lib/authenticated-api/authenticated-api.ts
mrliptontea: I made the change early when started working on this but I agree there's no need to expose it - I undid it.

lib/authenticated-api/authenticated-api.ts
mrliptontea: It's hard to tell - CircleCI pipeline logs expire and I couldn't see the exact issue we had then, but it's quite possible that was it.

PR #73, opened 2023-05-10
PLAT-6393 Lambda worker - allow fifo. Fix ecr command issue

malcyL:


test/integration/lambda-worker/lambda-worker.test.ts
```diff
@@ -1,4 +1,5 @@
 import { SQS } from "aws-sdk";
+import { v4 as uuidv4 } from "uuid";
```
malcyL, line 2: I've modified this integration test, to run against a fifo based lambda worker. This shows that it is properly working with fifo queues, both processing successes and DLQing failures.

There is only one integration test for the lambda-worker running against a deployed aws stack. And this modifies that test to be fifo based. I did consider adding another test, rather than modifying this one. So there would have been an integration test against a fifo and none fifo lambda-worker. 

But doping these deployments into AWS and running tests on them is at the expensive end of the testing pyramid. Both in terms of maintaining the tests and in running them in the build. I've decided not to double up the tests here, instead changing this one. There are infra unit tests which cover both the fifo and none fifo flavour of the construct.

However, if we think we should add another integration test (and fifo example as the tests run the examples) I will do.

kiyanwang (approved):
I think this looks ok, lets a get a green build and push forward

PR #80, opened 2023-09-15
chore: plat-6742 upgrade node in lambda-worker and authenticated-api

malcyL:
- https://github.com/talis/platform/issues/6727
- https://github.com/talis/platform/issues/6742

We have an epic to upgrade lambda's to Node 18. Multiple projects are using this library. The first of which to be picked up in Manifesto which uses both the lambda worker and the authenticated api constructs.

Unfortunately, CDK v1 has now been end of lifed and we can not upgrade past node 16 using it. 

This PR upgrades those two constructs to node 16.

### DIFF
--- a/lib/authenticated-api/authenticated-api-function.ts
+++ b/lib/authenticated-api/authenticated-api-function.ts
@@ -32,7 +32,7 @@ export class AuthenticatedApiFunction extends lambdaNode.NodejsFunction {
 
       // Enforce the following properties
       awsSdkConnectionReuse: true,
-      runtime: lambda.Runtime.NODEJS_14_X,
+      runtime: lambda.Runtime.NODEJS_16_X,
       timeout: props.timeout,
       securityGroups: props.securityGroups,
       vpc: props.vpc,
--- a/lib/authenticated-api/authenticated-api.ts
+++ b/lib/authenticated-api/authenticated-api.ts
@@ -167,7 +167,7 @@ export class AuthenticatedApi extends cdk.Construct {
         }),
 
         awsSdkConnectionReuse: true,
-        runtime: lambda.Runtime.NODEJS_14_X,
+        runtime: lambda.Runtime.NODEJS_16_X,
         timeout: authLambdaTimeout,
         securityGroups: props.securityGroups,
         vpc: props.vpc,
--- a/lib/lambda-worker/lambda-worker.ts
+++ b/lib/lambda-worker/lambda-worker.ts
@@ -294,7 +294,7 @@ export class LambdaWorker extends cdk.Construct {
 
       // Enforce the following properties
       awsSdkConnectionReuse: true,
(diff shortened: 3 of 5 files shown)


lib/authenticated-api/authenticated-api-function.ts
```diff
@@ -32,7 +32,7 @@ export class AuthenticatedApiFunction extends lambdaNode.NodejsFunction {
 
       // Enforce the following properties
       awsSdkConnectionReuse: true,
-      runtime: lambda.Runtime.NODEJS_14_X,
+      runtime: lambda.Runtime.NODEJS_16_X,
```
mrliptontea, line 35: Is the node 16 the limit because of enums? Not possible to use a string here?

malcyL: @mrliptontea TLDR: I don't believe so - no.

The type in the props is:

```
    /**
     * The runtime environment. Only runtimes of the Node.js family are
     * supported.
     *
     * @default Runtime.NODEJS_14_X
     */
    readonly runtime?: lambda.Runtime;
```

i.e. `lambda.Runtime` 

Runtime is a class:

```typescript
export declare class Runtime {
    /** A list of all known `Runtime`'s. */
    static readonly ALL: Runtime[];
    /**
     * The NodeJS runtime (nodejs)
     * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
     */
    static readonly NODEJS: Runtime;
    /**
     * The NodeJS 4.3 runtime (nodejs4.3)
     * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
     */
    static readonly NODEJS_4_3: Runtime;
    /**
     * The NodeJS 6.10 runtime (nodejs6.10)
     * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
     */
    static readonly NODEJS_6_10: Runtime;
    /**
     * The NodeJS 8.10 runtime (nodejs8.10)
     * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
     */
    static readonly NODEJS_8_10: Runtime;
    /**
     * The NodeJS 10.x runtime (nodejs10.x)
     * @deprecated Legacy runtime no longer supported by AWS Lambda. Migrate to the latest NodeJS runtime.
     */
    static readonly NODEJS_10_X: Runtime;
    /**
     * The NodeJS 12.x runtime (nodejs12.x)
     */
    static readonly NODEJS_12_X: Runtime;
    /**
     * The NodeJS 14.x runtime (nodejs14.x)
     */
    static readonly NODEJS_14_X: Runtime;
    /**
     * The NodeJS 16.x runtime (nodejs16.x)
     */
    static readonly NODEJS_16_X: Runtime;
```

If I try to assign a string - I get a typescript error:

image

Types for the win!

PR #84, opened 2023-10-26
feat: add alternate domain names to cdn site host construct

malcyL:
- https://techfromsage.atlassian.net/browse/PLT-62

# Problem

We have had outages to both smalti's internal UI and production elevate caused by the cloud front distribution deleting the alias for the live DNS name when a release is performed.

This is because we manually create the DNS entry and the cloud front distribution alias for it. We haven't included these in the cloud formation stack previously because we want to be able to follow a process of deploying a new stack alongside a live stack and switch to it when we are happy. We want to use DNS to do this in the same way we would when bringing up a new instance of a traditional web server. 

If the DNS entry or the cloud front distribution alias are contained in the cloud formation stack, a new stack with a different watermark can not be created alongside the live production stack. The DNS entry and the cloud front distribution alias can not be duplicated in another stack - the stack creation fails.

# Possible Solutions Discussed To Resolve This

These options are from this investigation ticket: https://github.com/talis/platform/issues/6436#issuecomment-1505271353

1. Have the deployment job in CircleCI re-add the alias it after doing deploy.
2. Include alternate domain name with the desired domain in the stack. Have a parameter to not include it, e.g. when we deploy into a new stack?
3. Remove CloudFront distribution from app stack (move to a separate stack or to terraform?). Treat Cloudfront Distribution like a DB, we provision once and only once.

Option 1 - Pro - minimal change to code. Con - Potential short downtime on every release. 
Option 2 - Pro - minimal change to code. Con - increase in complexity around deployments in hercules and individual projects
Option 3 - Pro - Deployment moved away from making any changes to Cloud Front. Con - this is a much bigger change

The decision was to go for Option 2. This PR implements that solution.

# Changes in This PR

Changes to the construct:

- adds a new property to CdnSiteHostingConstructProps called `aliasSubDomains`
- any entries in `aliasSubDomains` are added as aliases to the cloud front distribution
- when using CdnSiteHostingWithDnsConstruct, DNS entries are created for all aliases provided

A simple cdn site hosting construct example has been added to the examples section. This example:

- deploys a website with CdnSiteHostingConstructProps
- deploys a website with CdnSiteHostingWithDnsConstruct

The integration tests have been extended to run against this new sample project in the same way we have integration tests running against the other samples

The circle build has therefore been updated to deploy the new sample in parallel to it's deployments of the other samples.

The PR also adds a `Best Practices` section into the README to describe the imagined process for resolving this issue - changes are needed in both hercules and consuming projects.

# Smoke Testing

We have tested running through the envisaged process of bringing up a new stack and making it live alongside an existing live stack.

I was concerned about the final step - cloud formation adding in the new alias which has actually been added manually at the time of the DNS change.

However, this step did work and there are no issues with future releases after the switch.

Details here: https://techfromsage.atlassian.net/browse/PLT-62?focusedCommentId=110708


examples/simple-cdn-site-hosting-construct/test/simple-cdn-site-hosting-construct.test.ts
stephenrob, line 1: Do we need this test file in the example as its basically an empty file

stephenrob (approved):

examples/simple-cdn-site-hosting-construct/test/simple-cdn-site-hosting-construct.test.ts
malcyL: It's generated automatically as a place holder with the `cdk init --language typescript`.

It's sort of nice as a place holder - but I will delete it.

lib/cdn-site-hosting/cdn-site-hosting-construct.ts
```diff
@@ -44,7 +46,8 @@ export class CdnSiteHostingConstruct extends Construct {
         props.certificateArn,
       ),
       {
-        aliases: [siteDomain],
+        // aliases: [siteDomain],
+        aliases: aliases,
```
mrliptontea, line 50: I think I'd be happier with this:

```suggestion
        aliases: [siteDomain, ...aliases],
```

It makes it clear that the siteDomain is there and that it's the main one.

lib/cdn-site-hosting/utils.ts
```diff
@@ -2,3 +2,15 @@ export const getSiteDomain = (props: {
...
+
+export const getAliases = (props: {
+  domainName: string;
+  siteSubDomain: string;
+  aliasSubDomains?: string[];
+}): string[] => {
+  const aliases = props.aliasSubDomains
+    ? props.aliasSubDomains.map((alias) => `${alias}.${props.domainName}`)
+    : [];
+  aliases.push(getSiteDomain(props));
```
mrliptontea, line 14: Like I commented above, I don't think siteDomain should be here.

mrliptontea (approved):

junglebarry (approved):

PR #86, opened 2024-04-01
feat: lambda worker - support maxConcurrency (SqsEventSource)

nemanja-kovacevic-thinkit:
resolves: https://techfromsage.atlassian.net/browse/PLT-83

This PR adds support for maxConcurrency property in LambdaWorker (SqsEventSource)

### DIFF
--- a/lib/lambda-worker/lambda-worker.ts
+++ b/lib/lambda-worker/lambda-worker.ts
@@ -141,6 +141,7 @@ export class LambdaWorker extends Construct {
       new SqsEventSource(lambdaQueue, {
         enabled: props.lambdaProps.enableQueue ?? true,
         batchSize: 1,
+        maxConcurrency: props.lambdaProps.queueMaxConcurrency,
       }),
     );
     lambdaWorker.addEventSource(
--- a/examples/container-lambda-worker/lib/container-lambda-worker-stack.ts
+++ b/examples/container-lambda-worker/lib/container-lambda-worker-stack.ts
@@ -52,6 +52,7 @@ export class ContainerLambdaWorkerStack extends cdk.Stack {
       {
         name: `${prefix}container-lambda-worker`,
         lambdaProps: {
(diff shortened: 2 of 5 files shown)


kiyanwang (commented):
I think that's looking ok for exposing the new optional property, I think all this now needs is the tests updating, to verify that if you set the property the it really is exposed:

- https://github.com/talis/talis-cdk-constructs/blob/main/test/infra/lambda-worker/lambda-worker.test.ts
- https://github.com/talis/talis-cdk-constructs/tree/main/test/integration/lambda-worker

lib/lambda-worker/lambda-worker.ts
```diff
@@ -141,6 +141,7 @@ export class LambdaWorker extends Construct {
       new SqsEventSource(lambdaQueue, {
         enabled: props.lambdaProps.enableQueue ?? true,
         batchSize: 1,
+        maxConcurrency: props.lambdaProps.queueMaxConcurrency,
```
kiyanwang, line 144: what is the default value `maxConcurrency`  gets set to if you don't specify a value?
does it need a default value i.e. `maxConcurrency: props.lambdaProps.queueMaxConcurrency ?? <some default>`, or do we only set the property if a value for `queueMaxConcurrency` is set?

nemanja-kovacevic-thinkit: By default, maxConcurrency has no specific limit - uses up to maximum concurrency allowed by AWS.
I would add default value only if we internally know that there is some value that suits us for our use cases. Otherwise, I think that just forwarding a parameter and leaving default behavior is the best option.

kiyanwang: Lets not make this an optional property, lets make it a mandatory property. This is a breaking change, but will force us to enumerate the max concurrency for every use case, which is actually the right thing to do.

kiyanwang (approved):

PR #87, opened 2024-07-10
fix: update talis.io certificate

nemanja-kovacevic-thinkit:
[PLT-801](https://techfromsage.atlassian.net/browse/PLT-801)

Update talis.io certificate ARN to point to the new certificate.

[PLT-801]: https://techfromsage.atlassian.net/browse/PLT-801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stephenrob (approved):
We need to make sure down stream services that use the construct library are updated to the new version that includes this certificate post merge

PR #89, opened 2024-07-18
feat(ta-1105): add bundling options to authenticated api

mike-dean-talis:
## Description

* [TA-1105](https://techfromsage.atlassian.net/browse/TA-1105)

Whilst working on the new bookmarking extension API, we need to include `jsdom` in a lambda function. `jsdom` does not bundle correctly and requires extra options to help this along. Currently the `AuthenticatedApiFunction` does not expose the CDK bundling options.

## Change

* Expose the bundling options for an `AuthenticatedApiFunction`

[TA-1105]: https://techfromsage.atlassian.net/browse/TA-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/authenticated-api/authenticated-api-function-props.ts
+++ b/lib/authenticated-api/authenticated-api-function-props.ts
@@ -1,5 +1,6 @@
 import * as cdk from "aws-cdk-lib";
-import { BundlingOptions, aws_ec2 as ec2 } from "aws-cdk-lib";
+import { aws_ec2 as ec2 } from "aws-cdk-lib";
+import { BundlingOptions } from "aws-cdk-lib/aws-lambda-nodejs";
 
 export interface AuthenticatedApiFunctionProps {
   name: string;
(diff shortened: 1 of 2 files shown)


mrliptontea (approved):

lib/authenticated-api/authenticated-api-function-props.ts
```diff
@@ -11,4 +11,5 @@ export interface AuthenticatedApiFunctionProps {
   vpcSubnets?: ec2.SubnetSelection;
   securityGroups?: Array<ec2.ISecurityGroup>;
   memorySize?: number;
+  bundling?: BundlingOptions;
```
mikerudd, line 14: Is this type correct? I've just had a quick look at the [docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.BundlingOptions.html) and was expecting to see something related to `externalModules` & `nodeModules`.

mike-dean-talis: comment: Good catch. I had imported the wrong thing.

mikerudd (approved):

PR #93, opened 2025-03-14
feat(deps): move aws-cdk-lib, constructs to peerDependencies

mrliptontea:
Context: I believe this library was created from the wrong template:

```sh
cdk init app --language typescript
# instead of
cdk init lib --language typescript
```

The main difference is that `aws-cdk-lib` and `constructs` packages, in the library, should be in `peerDependencies` for upstream use, and `devDependencies` during development.

Additionally, other packages specified in `dependencies` can be moved:

- `axios`, `uuid` are only used in tests → `devDependencies`.
- `eslint`, `prettier` are only used in development and CI → `devDependencies`.
- `source-map-support` is only needed for apps, not libraries (again, wrong template) → removed entirely.
- `talis-node` depends on some old versions of libraries itself, therefore it's best to move it to its own `package.json`, and point NodejsFunction construct to it. Normally, Lambda dependencies are specified in the root `package.json`, but in this case, it's better to keep them separate.

This way all `dependencies` from this project are removed, with the exception of including a separate lock file for the authorizer function.


src/lambda/api/PersonaAuthorizer.ts
```diff
@@ -232,6 +232,7 @@ export class PersonaAuthorizer {
         persona_scheme: process.env["PERSONA_SCHEME"],
         persona_port: process.env["PERSONA_PORT"],
         persona_oauth_route: process.env["PERSONA_OAUTH_ROUTE"],
+        cert_background_refresh: false,
```
junglebarry, line 235: I'd really prefer this be kept separate - as per our guidance on separating functional and refactoring changes. If you have to roll this back, you lose the rest of the PR.

package.json
```diff
@@ -31,31 +31,31 @@
...
+    "constructs": "^10.0.0",
+    "conventional-changelog-conventionalcommits": "^5.0.0",
     "esbuild": "^0.19.3",
+    "eslint": "^8.57.1",
     "eslint-config-prettier": "^9.0.0",
+    "husky": "^8.0.3",
     "jest": "^29.6.4",
+    "prettier": "^3.5.3",
+    "semantic-release": "^18.0.1",
+    "talis-node": "github:talis/talis-node#v0.2.1",
```
mrliptontea, line 51: Should follow up - added this as a dev dependency to allow unit tests to pass.
The test in question is: https://github.com/techfromsage/talis-cdk-constructs/blob/f4d3cd9daffa70d54758a70fbaa98f0d53017673/test/unit/lambda/api/PersonaAuthorizer.test.ts#L1-L4

It imports `PersonaAuthorizer`, which requires talis-node, but the test only checks `pathMatch` function, which could easily be extracted.

Otherwise, `talis-node` package isn't used at all in this context.

junglebarry (approved):
One comment about mixing PRs, but otherwise LGTM.

junglebarry:
> Context: I believe this library was created from the wrong template:

Yeah, that sounds likely - mea culpa.

src/lambda/api/PersonaAuthorizer.ts
mrliptontea: Yeah I don't know what I was thinking trying to sneak that in here.

mrliptontea:
I've verified these changes:

```sh
# 1. Produce talis-cdk-constructs-0.1.0.tgz
npm run prepare
node_modules/.bin/npm pack # semantic-release install its own npm

# 2. Create a CDK app and install our constructs
cdk init app --language typescript
npm i ../../talis-cdk-constructs-0.1.0.tgz
npm i -D esbuild # if esbuild is not installed, CDK will use Docker to bundle

# 3. Add AuthenticatedApiFunction and AuthenticatedApi constructs to the stack

# 4. Synth to verify all assets are bundled correctly
npx cdk synth
```

Crevitus (approved):
Looks _great_!

PR #97, opened 2025-03-20
feat: allow custom NodejsFunction runtime

mrliptontea:
`nodejs18.x` Lambda runtime is going to be deprecated on Sep 1, 2025. To make upgrades easier, we should allow CDK apps to specify their own runtime version, rather than imposing one.

This PR makes constructs that use NodejsFunction construct expose additional props: `runtime`, `awsSdkConnectionReuse`, `depsLockFilePath`, `bundling`, `projectRoot`.

These are non-breaking changes. For backwards compatibility, we're keeping `runtime` defaulting to `nodejs18.x`, and `awsSdkConnectionReuse` defaulting to `true`. However, it should be noted, connection reuse setting has no real effect where AWS SDK v3 is used. The `AWS_NODEJS_CONNECTION_REUSE_ENABLED` environment variable does not exist in the AWS SDK for JavaScript v3 but clients [reuse connections by default](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-reusing-connections.html).

In the next PR, we should introduce breaking changes:

- Bump authoriser Lambda to `nodejs22.x` or `nodejs20.x`.
- Remove the default `true` from `awsSdkConnectionReuse`.

### DIFF
--- a/lib/authenticated-api/authenticated-api-function-props.ts
+++ b/lib/authenticated-api/authenticated-api-function-props.ts
@@ -1,8 +1,18 @@
 import * as cdk from "aws-cdk-lib";
 import { aws_ec2 as ec2 } from "aws-cdk-lib";
-import { BundlingOptions } from "aws-cdk-lib/aws-lambda-nodejs";
+import { aws_lambda_nodejs as lambda_nodejs } from "aws-cdk-lib";
 
-export interface AuthenticatedApiFunctionProps {
+export interface AuthenticatedApiFunctionProps
+  extends Pick<
+    lambda_nodejs.NodejsFunctionProps,
+    | "entry"
+    | "handler"
+    | "runtime"
+    | "awsSdkConnectionReuse"
+    | "depsLockFilePath"
+    | "bundling"
+    | "projectRoot"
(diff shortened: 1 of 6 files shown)


malcyL (changes requested):
> To make upgrades easier, we should allow CDK apps to specify their own runtime version, rather than imposing one.

Is this the correct approach?

One of the advantages we talked about when we created this library was that we could mandate certain things - minimum memory requirements on lambdas etc etc. The Node Runtime version was one of these.

Allowing this to be specified bu the application and not mandating it here means that when we get an update from AWS telling us that a runtime has been deprecated - we need to make changes in multiple projects.

If we want to update the node runtime - why not update it for everyone rather than individual projects? 

It also means we would have the authoriser running in production under different run times. We may see issues in one runtime and not in another. Ensuring they are all on the same runtime makes sense to me.

malcyL (changes requested):
There was a decision to mandate that `awsSdkConnectionReuse` is set to true.

Why do we want to make this configurable?

malcyL (approved):

PR #98, opened 2025-04-02
feat: adds a rest based api construct

malcyL:
- https://techfromsage.atlassian.net/browse/PLT-1355

This PR moves the Rest based Authenticated API Construct from the Metatron repo into the CDK Constructs Library.

This is an alternative to the existing AuthenticatedApi construct. The AuthenticatedApi construct is based on the ApiGateway V2 http api. This AuthenticatedRestApi is based on an ApiGateway V1 rest api.

While the AuthenticatedApi may remain the preferred construct, the migration of Metatron from Serverless to CDK did not migrate the service from the V1 Rest Api to the V2 HTTP Api.

Bringing the construct into the library means:
 
- we can re use the Rest Api construct if we choose
- we can consoldate the authentication between the two constructs

#### Authentication

The new construct adds a new authentication lambda `rest-api/authorizer.ts` alongside the existing `api/authorizer.ts`.

I did look at consolidating these into one. However, both the input event, and the response are different between the two. In the end I felt it was better to keep them separate, especially considering our longer term aim here is to remove this old authentication method. Spending time on it, when the complexity grew, seemed like the wrong call. But we could pick this up if we choose: PLT-1355-RestBasedAPI_consolidate_auth

Metatron also does not uses scopes in its authentication. While token validation is implemented, the scope validation implemented in AuthenticatedApi has not been implemented in this PR. Again, it's not simple and the time may be better spent on replacing this auth mechanism.

However, having these two constructs together in this library puts us in a better place to consolidate the authentication when we do schedule work to replace it.


examples/simple-authenticated-api/lib/simple-authenticated-api-stack.ts
```diff
@@ -27,7 +27,7 @@ export class SimpleAuthenticatedApiStack extends cdk.Stack {
     // Can we pull in this alarm which is defined in terraform as an example of how to do that
     const alarmTopic = new sns.Topic(
       this,
-      `${prefix}simple-lambda-Worker-alarm`,
+      `${prefix}simple-authenticated-api-alarm`,
```
malcyL, line 30: Correcting historic copy paste error.

README.md
```diff
@@ -22,12 +22,19 @@ Builds are conducted by CircleCI, and upon successful build of the `main` branch
...
   - Details [here](/examples/simple-lambda-worker/README.md).
 
-- `AuthenticatedApi`
+- `AuthenticatedApi`(http api)
 
   - An API Gateway (v2)
   - Built in optional Persona authentication on routes
   - Details [here](/examples/simple-authenticated-api/README.md).
 
+- `AuthenticatedResApi`(rest api)
```
junglebarry, line 31:
```suggestion
- `AuthenticatedRestApi`(rest api)
```

?

examples/simple-authenticated-rest-api/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "commonjs",
+    "lib": ["es2020", "dom"],
```
junglebarry, line 5: do you need `dom`? This doesn't look like it uses a DOM.

lib/authenticated-rest-api/authenticated-rest-api.ts
```diff
@@ -0,0 +1,272 @@
...
+    super(scope, id);
+
+    if (
+      (props.domainName && !props.certificateArn) ||
+      (!props.domainName && props.certificateArn)
+    ) {
+      cdk.Annotations.of(scope).addError(
+        `To use a custom domain name both certificateArn and domainName must be specified`,
+      );
+    }
```
junglebarry, line 38: The conditional looks quite complicated when compared to the statement in the message. Are they both mandatory? Or is it both or neither?

lib/authenticated-rest-api/authenticated-rest-api.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      new cdk.CfnOutput(this, "apiGatewayLogGroup", {
+        exportName: `${apiName}-accessLogGroup`,
+        value: logGroup.logGroupName,
+      });
+    }
+
+    this.alarmAction = new cloudwatchActions.SnsAction(props.alarmTopic);
+
+    // TODO: Validation of scopes is not yet implemented.
```
junglebarry, line 84: - [ ] Resolve

lib/authenticated-rest-api/authenticated-rest-api.ts
```diff
@@ -0,0 +1,272 @@
...
+      .with({ statistic: "average", period: cdk.Duration.minutes(1) });
+
+    const routeLatencyAlarm = new cloudwatch.Alarm(
+      this,
+      `${apiName}-latency-alarm`,
+      {
+        alarmName: `${apiName}-latency-alarm`,
+        alarmDescription: `Alarm if latency on api ${props.prefix}${
+          props.name
+        } exceeds ${latencyThreshold.toMilliseconds()} milliseconds`,
```
junglebarry, line 164:
```suggestion
        alarmDescription: `Alarm if latency on api ${apiName} exceeds ${latencyThreshold.toMilliseconds()} milliseconds`,
```

lib/authenticated-rest-api/authenticated-rest-api.ts
```diff
@@ -0,0 +1,272 @@
...
+      deployOptions: {
+        stageName: props.stageName,
+      },
+      ...(props.corsDomain && {
+        corsPreflight: {
+          allowHeaders: ["*"],
+          allowMethods: ["ANY"],
+          allowCredentials: true,
+          allowOrigins: props.corsDomain,
+        },
```
junglebarry, line 59: Do we really want this to be so permissive?

lib/authenticated-rest-api/authenticated-rest-api.ts
```diff
@@ -0,0 +1,272 @@
...
+  addResource(
+    apiName: string,
+    resourceProps: ResourceProps,
+    authorizer: apigateway.IAuthorizer,
+    parent?: apigateway.Resource,
+  ) {
+    let actualParent = this.restApi.root;
+    if (parent) {
+      actualParent = parent;
+    }
```
junglebarry, line 189:
```suggestion
    const actualParent = parent ?? this.restApi.root;
```

If it doesn't change, `const` removes the cognitive load of having to read the code to find that out.

src/lambda/rest-api/PersonaAuthorizer.js
junglebarry, line 1: Is this autogenerated? Should it be ignored so it can be built instead?

src/lambda/rest-api/PersonaAuthorizer.d.ts
junglebarry, line 1: As below - ignored and built?

src/lambda/rest-api/PersonaAuthorizer.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    this.personaClient = undefined;
+  }
+
+  async handle() {
+    console.log("Received event", this.event);
+
+    if (
+      !this.event?.authorizationToken ||
+      this.event.authorizationToken == null
```
junglebarry, line 73:
```suggestion
      !this.event?.authorizationToken
```

`null` is falsy.

src/lambda/rest-api/PersonaAuthorizer.ts
```diff
@@ -0,0 +1,310 @@
...
+      const authPolicy = this.buildAuthPolicy(
+        principalId,
+        parsedMethodArn,
+        true,
+      );
+      return this.context.succeed(authPolicy);
+    } catch (err) {
+      console.log("token validation failed", err);
+
+      // In the case of suceess - the principal id is coming from the
```
junglebarry, line 115:
```suggestion
      // In the case of success - the principal id is coming from the
```

src/lambda/rest-api/PersonaAuthorizer.ts
```diff
@@ -0,0 +1,310 @@
...
+      // decoded token. We don't have it here for the case of an invalid token.
+      // Leaving the proncipal id blank in the auth policy for deny for now.
+      // But could this be found?
```
junglebarry, line 118: What does this mean?

test/integration/authenticated-rest-api/authenticated-rest-api.test.ts
```diff
@@ -0,0 +1,136 @@
+import axios from "axios";
+import { APIGateway } from "aws-sdk";
+
+const api = new APIGateway();
+
+const TALIS_CDK_AUTH_API_VALID_CLIENT =
+  process.env.TALIS_CDK_AUTH_API_VALID_CLIENT ?? "";
+const TALIS_CDK_AUTH_API_VALID_SECRET =
+  process.env.TALIS_CDK_AUTH_API_VALID_SECRET ?? "";
```
junglebarry, line 9: Do we want empty string literals here? What do they mean?

src/lambda/rest-api/PersonaAuthorizer.ts
```diff
@@ -0,0 +1,310 @@
...
+    parsedMethodArn: ParsedArn,
+    allow: boolean,
+  ) {
+    const policy = new AuthPolicy(
+      principalId,
+      parsedMethodArn.awsAccountId,
+      parsedMethodArn.apiOptions,
+    );
+
+    // const versionedResourcePath = `/${parsedMethodArn.apiVersion}${parsedMethodArn.resourcePath}`;
```
junglebarry, line 295: Remove?

src/lambda/rest-api/PersonaAuthorizer.ts
```diff
@@ -0,0 +1,310 @@
...
+   *   resourcePath: string,
+   *   apiOptions: {
+   *     region: string,
+   *     restApiId: string,
+   *     stage: string
+   *   },
+   *   awsAccountId: string
+   *   }}
+   */
+  parseMethodArn(methodArn: string): ParsedArn {
```
junglebarry, line 170: Do we have unit-tests covering this? It feels fragile.

junglebarry (commented):
Some minor trivia and typos, I suspect

examples/simple-authenticated-rest-api/tsconfig.json
malcyL: This file is auto generated. It's the same as in all the other auto generated `tsconfig.json` files in these projects: 

- https://github.com/techfromsage/talis-cdk-constructs/blob/main/examples/simple-authenticated-api/tsconfig.json#L5
- https://github.com/techfromsage/talis-cdk-constructs/blob/main/examples/simple-lambda-worker/tsconfig.json#L5
- https://github.com/techfromsage/talis-cdk-constructs/blob/main/examples/simple-talis-cdk-stack/tsconfig.json#L5

I'm not sure about changing the auto generated files unless there is a strong reason to do so.

lib/authenticated-rest-api/authenticated-rest-api.ts
malcyL: Neither are mandatory, but if one is specified, they both have to be.

lib/authenticated-rest-api/authenticated-rest-api.ts
malcyL: I put in the description of the PR, my intention not to resolve this here. Are you saying this has to be resolved before this PR can be merged?

src/lambda/rest-api/PersonaAuthorizer.ts
malcyL: Great question. Probably one for CC - I guess this has always been here. Happy to delete the comment.

test/integration/authenticated-rest-api/authenticated-rest-api.test.ts
malcyL: When the env vars are not set, they are empty and the tests fail.

Happy to change this, but at some point we have to deal with the type of `process.env.TALIS_...` being `string | undefined`. Where these constants are being used, the type is `string`.

Is there a preferred way of reading env vars and using them as strings?

What is your suggestion here?

lib/authenticated-rest-api/authenticated-rest-api.ts
malcyL: Maybe not. They are both mandatory, so this can be removed.

lib/authenticated-rest-api/authenticated-rest-api.ts
malcyL: This matches what we have in the existing construct:

- https://github.com/techfromsage/talis-cdk-constructs/blob/main/lib/authenticated-api/authenticated-api.ts#L52

```javascript
    const apiGatewayProps: apigatewayv2.HttpApiProps = {
      apiName: apiName,
      defaultDomainMapping: { domainName: this.domainName },
      ...(props.corsDomain && {
        corsPreflight: {
          allowHeaders: ["*"],
          allowMethods: [apigatewayv2.CorsHttpMethod.ANY],
          allowCredentials: props.corsAllowCredentials ?? true,
          allowOrigins: props.corsDomain,
        },
      }),
    };
```

And this is how metatron is currently configured. I think that is the only ApiGateway we have not using the existing construct.

So it is certainly how things have been upto now. I'm not against changing it. But if we want to change it - could we move it into another story/pr and make the change in the other construct too. We wouldthen need to roll the update out to everywhere which is using it.

lib/authenticated-rest-api/authenticated-rest-api.ts
junglebarry: We have a long-standing policy of not leaving TODOs or commented code in our codecases. I'd be fine with you simply removing the TODO.

test/integration/authenticated-rest-api/authenticated-rest-api.test.ts
junglebarry: I'm generally not a fan of coercing variables to silence type-system feedback. I'd typically view empty string literals as a code smell.

You could validate the environment first - save yourself even running the tests - a basic condition on the line below, or a validation function, perhaps. For something a little more comprehensive, you could do something like Grzeg's validated env in Content Linking:

https://github.com/techfromsage/content-linking-monorepo/blob/main/apps/api/src/config/config.ts

Or you could carry on with empty string literals - the test still fails. YMMV.

junglebarry (approved):

PR #100, opened 2025-04-03
fix: replace talis node as dependency

malcyL:
- https://techfromsage.atlassian.net/browse/PLT-1355

In PR https://github.com/techfromsage/talis-cdk-constructs/pull/93 we moved `talis-node` to be a dev dependency.

It is not a dev dependency - it is used by the authorizer class in the API Gateway construct. Removing it results in the following deployment errors:

image

This PR replaces it as a dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,9 @@
     "": {
       "name": "talis-cdk-constructs",
       "version": "0.1.0",
+      "dependencies": {
+        "talis-node": "github:talis/talis-node#v0.2.1"
+      },
       "devDependencies": {
         "@commitlint/cli": "^17.6.1",
         "@commitlint/config-conventional": "^17.6.1",
@@ -27,7 +30,6 @@
         "jest": "^29.6.4",
         "prettier": "^3.5.3",
         "semantic-release": "^18.0.1",
-        "talis-node": "github:talis/talis-node#v0.2.1",
         "ts-jest": "^29.1.1",
         "ts-node": "^10.9.1",
         "typescript": "~5.2.2",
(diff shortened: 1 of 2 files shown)


mrliptontea (changes requested):
We don't want that package as a dependency of the library, which is what #93 was all about - to not impose really old packages via `talis-node` on apps using this library.

The reason `talis-node` is in `devDependencies` at all, is (are) PersonaAuthorizer unit test(s), which by itself is something that we could refactor, as these tests only check helper functions, not the authoriser itself, so these functions could be moved to separate files thus allowing to remove `talis-node` from `devDependencies`.

`talis-node` is a dependency of a Lambda, and one that ideally we'd get rid of, so instead there's a separate package.json and npm-shrinkwrap.json that are used for bundling.

From this package.json we create npm-shrinkwrap.json:

https://github.com/techfromsage/talis-cdk-constructs/blob/0250cd445d43937396bea2709438d756467a2205/src/package.json#L1-L6

We use shrinkwrap instead of regular lock because [it can be published with a package](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson). Then, we point the NodeJsFunction to the non-default lock file, plus we use hooks to ensure dependencies are installed automatically for bundling:

https://github.com/techfromsage/talis-cdk-constructs/blob/0250cd445d43937396bea2709438d756467a2205/lib/authenticated-api/authenticated-api.ts#L140-L151

So instead of re-adding that unwanted library, in `authenticated-rest-api.ts`, set `depsLockFilePath` and use bundling `commandHooks` to run `npm ci`.

